PR #30, opened 2023-09-29
29 Fix NA buyer/seller name index error

### LINKED ISSUE #29: Condo flagging bug
When running flagging on pre-2017 data, we get an `IndexError` in the `name_selector()` function. It could have something to do with the NA values in buyer and seller name.

wagnerlmichael:
The code now runs with the entire 2014- sales query.  I believe the reason that it was not running was because `name_selector()` didn't handle the case where `split_logic()` looked at NA values in buyer or seller name.

`name_selector()` takes the return value of `split_logic()` as an argument in `clean_id()`. Specifically this happens in these two lines:

``` 
df["sv_buyer_id"] = df.apply(clean_id, args=("sv_buyer",), axis=1)
df["sv_seller_id"] = df.apply(clean_id, args=("sv_seller",), axis=1)
```
The fix here essentially converts the NA values to an "Empty Name" string value.

Before, there was an out of index error, since the NA value was being indexed.

Closes #29

### DIFF
--- a/glue/flagging_script_glue/flagging_e894e5.py
+++ b/glue/flagging_script_glue/flagging_e894e5.py
@@ -1,6 +1,6 @@
 """
 This file contains all necessary functions to create a DataFrame ready to use for
-non-arms length transaction detection using statistical and heurstic methods.
+non-arms length transaction detection using statistical and heuristic methods.
 """
 
 import re
@@ -546,9 +546,10 @@ def get_sale_counts(dups: pd.DataFrame) -> pd.DataFrame:
         df (pd.DataFrame): pandsa dataframe4
     """
     v_counts = (
-        dups.pin.value_counts().reset_index()
-        # .rename(columns={"count": "sv_sale_dup_counts"})
-        .rename(columns={"index": "pin", "pin": "sv_sale_dup_counts"})
+        dups.pin.value_counts()
+        .reset_index()
+        .rename(columns={"count": "sv_sale_dup_counts"})
+        # .rename(columns={"index": "pin", "pin": "sv_sale_dup_counts"})
     )
 
     dups = pd.merge(dups, v_counts)
@@ -884,6 +885,17 @@ def get_id(row: pd.Series, col: str) -> str:
     column = col + "_name"
     words = str(row[column]).lower()
 
+    # Check for missing values first
+    if pd.isnull(row[column]) or words in [
+        "none",
+        "nan",
+        "unknown",
+        "missing seller name",
+        "missing buyer name",
+    ]:
+        id = "Empty Name"
+        return id
+
     words = re.sub(r" amp ", "", words)
     words = re.sub(" +", " ", words)
 
@@ -1051,9 +1063,8 @@ def split_logic(words: str):
         words (str): cleaned str from get_id
     Outputs:
         'Empty Name' if string is empty
-        tokens (list): lsit of tokens in string from split
+        tokens (list): list of tokens in string from split
     """
-
     words = re.sub(" +", " ", words)
 
     if words.isspace() or re.search(r"^[.]*$", words) or words == "Empty Name":
@@ -1063,7 +1074,7 @@ def split_logic(words: str):
 
     _and = re.search(
         r"\b and\b|\b an$\b|\b a$\b|f k a|\bfka\b| n k a|\bnka\b|"
-        r"\b aka\b|a k a|\b kna\b|k n a| f k$|n k$|a k$|\b not\b| married",
+        r"\b aka\b|a k a(?=\\s|$)|\b kna\b|k n a| f k$|n k$|a k$|\b not\b| married",
         words,
     )
 
@@ -1085,22 +1096,14 @@ def name_selector(tokens) -> str:
         'Empty Name' if name is empty.
         id (str): identified last name
     """
-    if not tokens:
+
+    if tokens == "Empty Name" or tokens == []:
         return "Empty Name"
-    if tokens == "Empty Name":
-        return tokens
-    # Ex: John Smith Jr
+
     if tokens[-1] in ["jr", "sr", "ii", "iii", "iv", "v"]:
         tokens = tokens[:-1]
-    # Ex: John Smith
-    elif len(tokens) == 2:
-        id = tokens[1]
-    # John George Smith
-    if len(tokens) == 3:
-        id = tokens[2]
-    # John George Theodore Smith
-    else:
-        id = tokens[-1]
+
+    id = tokens[-1]
 
     return id
 
@@ -1165,7 +1168,7 @@ def clean_id(row: pd.Series, col: str) -> str:
     """
     Cleans id field after get_role() by removing role.
     Inputs:
-        row: from padnas dataframe
+        row: from pandas dataframe
         col (str): column to process. 'seller' or 'buyer'
     Outputs:
         words (str): seller/buyer id without role.
@@ -1207,23 +1210,6 @@ def create_judicial_flag(df: pd.DataFrame) -> pd.DataFrame:
     )
 
     return df
-
-
-def create_match_flag(row: pd.Series) -> str:
-    """
-    Creates a column that says whether the buyer/seller id match.
-    Meant for apply().
-    Inputs:
-        row: from dataframe
-    Outputs:
-        value (str): whether the buyer and seller ID match
-    """
-    if row["buyer_id"] == row["seller_id"] and row["buyer_id"] != "Empty Name":
-        value = "Buyer ID and Seller ID match"
-    else:
-        value = "No match"
-
-    return value
 
 
 def create_name_match(row: pd.Series) -> str:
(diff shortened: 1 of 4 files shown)


glue/flagging_script_glue/flagging_e894e5.py
```diff
@@ -1085,13 +1085,15 @@ def name_selector(tokens) -> str:
...
         return tokens
     # Ex: John Smith Jr
     if tokens[-1] in ["jr", "sr", "ii", "iii", "iv", "v"]:
         tokens = tokens[:-1]
     # Ex: John Smith
-    if len(tokens) == 2:
+    elif len(tokens) == 2:
         id = tokens[1]
     # John George Smith
     if len(tokens) == 3:
```
dfsnow, line 1099: **question (non-blocking):** Shouldn't this line also be `elif` given the proceeding line? Think carefully about the control flow here.

glue/flagging_script_glue/flagging_e894e5.py
```diff
@@ -546,9 +546,9 @@ def get_sale_counts(dups: pd.DataFrame) -> pd.DataFrame:
         df (pd.DataFrame): pandsa dataframe4
     """
     v_counts = (
-        dups.pin.value_counts().reset_index()
-        # .rename(columns={"count": "sv_sale_dup_counts"})
-        .rename(columns={"index": "pin", "pin": "sv_sale_dup_counts"})
+        dups.pin.value_counts()
+        .reset_index()
+        .rename(columns={"count": "sv_sale_dup_counts"})
```
wagnerlmichael, line 551: This is changed because we have standardized pandas versions in glue and on local.

glue/flagging_script_glue/flagging_e894e5.py
```diff
@@ -884,6 +884,17 @@ def get_id(row: pd.Series, col: str) -> str:
...
+    if pd.isnull(row[column]) or words in [
+        "none",
+        "nan",
+        "unknown",
+        "missing seller name",
+        "missing buyer name",
+    ]:
+        id = "Empty Name"
+        return id
+
```
wagnerlmichael, line 897: The prior flagging script was erroneously matching values with different string representations of NA. Since there are null values, the `words = str(row[column]).lower()` directly above this code chunk was converting those to `none` or `nan`, which returned family sales matches. 

There were also other already present values such as `UNKOWN`, `MISSING SELLER NAME`, and `MISSING BUYER NAME`, which were returning matches. We decide here to manually change these values to `Empty Name`, using the existing strategy.

glue/flagging_script_glue/flagging_e894e5.py
```diff
@@ -1207,23 +1211,6 @@ def create_judicial_flag(df: pd.DataFrame) -> pd.DataFrame:
...
-        value (str): whether the buyer and seller ID match
-    """
-    if row["buyer_id"] == row["seller_id"] and row["buyer_id"] != "Empty Name":
-        value = "Buyer ID and Seller ID match"
-    else:
-        value = "No match"
-
-    return value
-
-
```
wagnerlmichael, line 1226: This was an unused function which was an exact duplicate of the `df_name_match` function.

glue/flagging_script_glue/flagging_e894e5.py
```diff
@@ -1063,7 +1073,7 @@ def split_logic(words: str):
 
     _and = re.search(
         r"\b and\b|\b an$\b|\b a$\b|f k a|\bfka\b| n k a|\bnka\b|"
-        r"\b aka\b|a k a|\b kna\b|k n a| f k$|n k$|a k$|\b not\b| married",
+        r"\b aka\b|a k a(?=\\s|$)|\b kna\b|k n a| f k$|n k$|a k$|\b not\b| married",
```
wagnerlmichael, line 1076: (No real names used in this comment) This small regex fix was what actually fixed the error. After this regex, there is a split on the match which take the left element. In one case a name had the structure like `A. K. allison`, so it split on this name returning an empty string. The old match structure also hit names like:

- Samantha K. Adams
- Augustina K. Anthony

We don't want these matches, so I added a lookahead to see if there was another letter after the final `'a'`. After manually reviewing results, this improved matching.

glue/flagging_script_glue/flagging_e894e5.py
```diff
@@ -1085,20 +1095,14 @@ def name_selector(tokens) -> str:
         'Empty Name' if name is empty.
         id (str): identified last name
     """
-    if tokens == "Empty Name":
-        return tokens
-    # Ex: John Smith Jr
+
+    if tokens == "Empty Name" or tokens == []:
+        return "Empty Name"
+
```
wagnerlmichael, line 1101: This change prevents the same error from occurring.

In order to make sure the index error from returning an empty list doesn't cause the whole job to break, I changed this such that if it happens again, it will just read it as `'Empty Name'`, instead of a null list value.

glue/flagging_script_glue/flagging_e894e5.py
```diff
@@ -1085,20 +1095,14 @@ def name_selector(tokens) -> str:
...
-    if len(tokens) == 2:
-        id = tokens[1]
-    # John George Smith
-    if len(tokens) == 3:
-        id = tokens[2]
-    # John George Theodore Smith
-    else:
-        id = tokens[-1]
+
+    id = tokens[-1]
```
wagnerlmichael, line 1105: We simplify the logic here as the prior code in this function was redundant (grabbing last element of list)

glue/flagging_script_glue/flagging_e894e5.py
dfsnow: **praise:** Nice catch @wagnerlmichael.

**Question:** I assume that the name matching down the line discards any matches on `"Empty Name"`, is that correct?

glue/flagging_script_glue/flagging_e894e5.py
dfsnow: **praise:** Nice find again @wagnerlmichael. This kind of stuff is super tricky to debug.

glue/flagging_script_glue/flagging_e894e5.py
dfsnow: **request:** Can you add more context about what the error is here, so that I (and future parties) can understand exactly what was breaking and what this will fix?

glue/flagging_script_glue/flagging_e894e5.py
```diff
@@ -1085,22 +1095,14 @@ def name_selector(tokens) -> str:
...
-    if not tokens:
+
+    if tokens == "Empty Name" or tokens == []:
         return "Empty Name"
-    if tokens == "Empty Name":
-        return tokens
-    # Ex: John Smith Jr
+
     if tokens[-1] in ["jr", "sr", "ii", "iii", "iv", "v"]:
         tokens = tokens[:-1]
```
dfsnow, line 1103: **nitpick:** What if the last two elements match one of these? i.e. James Kirk Jr. III.

glue/flagging_script_glue/flagging_e894e5.py
wagnerlmichael: Yes, that is correct

manual_flagging/src/flagging_rolling.py
```diff
@@ -1079,27 +1088,25 @@ def split_logic(words: str):
...
-    if len(tokens) == 2:
-        id = tokens[1]
-    # John George Smith
-    if len(tokens) == 3:
-        id = tokens[2]
-    # John George Theodore Smith
-    else:
-        id = tokens[-1]
+        if not tokens:  # Avoids IndexError if all tokens are removed.
+            return "Empty Name"
```
wagnerlmichael, line 1107: This fixes the problem of multiple suffixes like `'Bob jr iii'`

glue/flagging_script_glue/flagging_e894e5.py
wagnerlmichael: Sure. So in `split_logic()`, there are various scenarios where text is split on a certain text match. The error that was happening happened to do with a name like `'A. K. Arnold'`. In the old logic, the `'a k a'` was detected as a match, and split on this name. 

In that part of `split_logic()`, we take the content of whatever was to the _left_  of the split. In the case of  `'A. K. Arnold'`, the left part of the string after the split would have been `null`, or an empty string. Interestingly, this never happened until one problematic name in the condo data came through. So the output `split_logic()` was an empty list, in this case.

Then, the output from `split_logic()` feeds into `name_selector()`, the input was an empty list, and the code from name selector attempts to index a value in the empty list, leading to an index error.

The fix then, just manually adjusts the empty list to an `'Empty Name`' value, which will just lead to a non-matched name.

Let me know if this makes sense @dfsnow

dfsnow: Checks out, thanks for the extra context!

manual_flagging/src/flagging_rolling.py
dfsnow: I appreciate the handling for the unfortunate soul named `Jr. Jr. Jr.`.

dfsnow (approved):
Looks good @wagnerlmichael! I'd ask Jean about merge order here since we might want to wait until the terraform merge is done.

PR #31, opened 2023-10-02
Use Terraform to manage Glue job deployments

### LINKED ISSUE #11: Refactor Glue setup for public repo + IaC
Currently, this repo is set to private and is a draft product. However, in the near future, we should make it public, since it will be a critical part of our data prep for modeling and reporting.

As such, we need to clean this repo of any secrets used during during development, such as roles or bucket names. We should also use this as a test case for the principles outlined in the [dbt/Glue design doc](https://github.com/ccao-data/data-architecture/blob/master/documentation/design-docs/data-catalog.md) and use Terraform (probably OpenTF now) to generate/tear down the Glue job.

jeancochrane:
⚠️ This PR deletes the `flagging_hash` column from the flagged sale output, so once it gets merged we'll need to recreate and recrawl the prod data to reflect the schema change. ⚠️ 

This PR replaces the built-in Glue deployment integration with a custom Terraform config that is capable of deploying staging jobs in addition to the production job. This config is used alongside a set of GitHub Actions workflows to implement a proper CI/CD pipeline for this repo, including:

* Deploys of staging resources on all pushes to pull request branches
* Deploys of production resources on all pushes to the main branch
* Automatic replacement of the flagging script without requiring the legacy `hash.sh` script to be run manually
* Automatic cleanup of staging resources when pull requests are closed

For an example of a successful deployment job (`deploy-terraform`), see [this workflow run](https://github.com/ccao-data/model-sales-val/actions/runs/6386066144/job/17332084284). For an example of a successful cleanup job (`cleanup-terraform`), see [this workflow run](https://github.com/ccao-data/model-sales-val/actions/runs/6386009283/job/17332052885).

Closes #11.


.github/workflows/cleanup-terraform.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+      - name: Setup Terraform
+        uses: ./.github/actions/setup-terraform
+        with:
+          role-to-assume: ${{ secrets.AWS_IAM_ROLE_TO_ASSUME_ARN }}
+          terraform-iam-role-arn: ${{ secrets.TERRAFORM_IAM_ROLE_ARN }}
+          aws-account-id: ${{ secrets.AWS_ACCOUNT_ID }}
+
+      - name: Delete resources using Terraform
+        run: terraform destroy -auto-approve
```
jeancochrane, line 29: The [`destroy` command](https://developer.hashicorp.com/terraform/cli/commands/destroy) is a shortcut for deleting all resources in the current workspace, in this case the workspace for the PR that is being closed. The [`-auto-approve` flag](https://developer.hashicorp.com/terraform/cli/commands/apply#automatic-plan-mode) overrides Terraform's default behavior to indicate that we do not want Terraform to wait for `y`/`n` prompt input before applying changes.

.github/workflows/cleanup-terraform.yaml
```diff
@@ -0,0 +1,30 @@
+---
+name: cleanup-terraform
```
jeancochrane, line 2: Similar to [our dbt setup](https://github.com/ccao-data/data-architecture/blob/a5684e69b6534c02a9e13a01497ed9e3236d0d46/.github/workflows/cleanup_dbt_resources.yaml#L1), this workflow destroys all staging resources when a PR is closed. For an example of a successful run of this workflow, [see here](https://github.com/ccao-data/model-sales-val/actions/runs/6386009283/job/17332052885).

.github/workflows/deploy-terraform.yaml
```diff
@@ -0,0 +1,35 @@
...
+          role-to-assume: ${{ secrets.AWS_IAM_ROLE_TO_ASSUME_ARN }}
+          terraform-iam-role-arn: ${{ secrets.TERRAFORM_IAM_ROLE_ARN }}
+          aws-account-id: ${{ secrets.AWS_ACCOUNT_ID }}
+
+      - name: Validate Terraform config
+        run: terraform validate
+        shell: bash
+
+      - name: Apply Terraform changes
+        run: terraform apply -auto-approve
```
jeancochrane, line 34: The [`apply` command](https://developer.hashicorp.com/terraform/cli/commands/apply) first diffs the local Terraform config defined in `main.tf` against the state of remote resources in AWS, and applies any necessary changes based on that diff (including creating, updating, or deleting resources)

.github/workflows/deploy-terraform.yaml
```diff
@@ -0,0 +1,35 @@
+---
+name: deploy-terraform
```
jeancochrane, line 2: For an example of a successful run of this workflow, [see here](https://github.com/ccao-data/model-sales-val/actions/runs/6386066144/job/17332084284).

glue/sales_val_flagging.py
```diff
@@ -424,7 +424,7 @@ def get_parameter_df(
     return df_parameters
 
 
-def get_metadata_df(run_id, timestamp, run_type, commit_sha, flagging_hash=""):
```
jeancochrane, line 427: The flagging script is now re-deployed with every commit, so we no longer need to store a reference to its hash; instead, the `commit_sha` should uniquely identify the state of the flagging script at the time that the script was deployed.

One question on this: Do we need to alter the data model at all to remove references to `flagging_hash`?

main.tf
```diff
@@ -0,0 +1,166 @@
...
+    }
+  }
+
+  required_version = ">= 1.5.7"
+
+  backend "s3" {
+    bucket               = "ccao-terraform-state-us-east-1"
+    key                  = "workspaces/default/terraform.tfstate"
+    region               = "us-east-1"
+    workspace_key_prefix = "workspaces"
```
jeancochrane, line 15: This block defines the [backend](https://developer.hashicorp.com/terraform/language/settings/backends/configuration) that Terraform will use to manage state. The `key` attribute defines the location of the default workspace, but we don't use that workspace at all in our GitHub Actions config, so workspace state files will instead follow the naming scheme set by `workspace_key_prefix` and will be stored under `s3://ccao-terraform-state-us-east-1/workspaces/<workspace>/terraform.tfstate`.

main.tf
```diff
@@ -0,0 +1,166 @@
...
+  }
+}
+
+provider "aws" {
+  region = "us-east-1"
+}
+
+locals {
```
jeancochrane, line 23: [Locals](https://developer.hashicorp.com/terraform/language/values/locals) are essentially reusable expressions. We use them below to define constants, including some constants that are defined differently depending on the workspace.

main.tf
```diff
@@ -0,0 +1,166 @@
...
+  s3_prefix                = "scripts/sales-val"
+  s3_bucket_data_warehouse = terraform.workspace == "prod" ? "ccao-data-warehouse-us-east-1" : aws_s3_bucket.data_warehouse[0].id
+  s3_bucket_glue_assets    = terraform.workspace == "prod" ? "ccao-glue-assets-us-east-1" : aws_s3_bucket.glue_assets[0].id
```
jeancochrane, line 26: Production buckets are shared by other Glue jobs, and so they aren't managed (read: created, updated, or destroyed) by this Terraform config, but will instead by managed by our upcoming org-wide config.

main.tf
```diff
@@ -0,0 +1,166 @@
...
+  glue_job_name            = "sales_val_flagging${terraform.workspace == "prod" ? "" : "_${terraform.workspace}"}"
+  glue_crawler_name        = "ccao-data-warehouse-sale-crawler${terraform.workspace == "prod" ? "" : "-${terraform.workspace}"}"
```
jeancochrane, line 28: Jobs and crawlers will have name suffixes on staging environments that should make them easier to distinguish from their prod equivalents. If you have alternate suggestions for namespacing schemes, let me know!

main.tf
```diff
@@ -0,0 +1,166 @@
...
+  description = <<EOF
+    "Hash of the current git commit, to use for versioning the model. "
+    "Generate with `git rev-parse --short HEAD`"
+  EOF
+  nullable    = false
+}
+
+resource "aws_s3_bucket" "glue_assets" {
+  # Prod buckets are managed outside this config
+  count         = terraform.workspace == "prod" ? 0 : 1
```
jeancochrane, line 49: It is perhaps [a miserable hack](https://tomharrisonjr.com/terraform-count-is-a-miserable-hack-d58a6ffbf422), but setting the [`count`](https://developer.hashicorp.com/terraform/language/meta-arguments/count) of a resource to `0` on specific environments is currently the best-supported way of [deploying different resources in different environments](https://medium.com/johnveldboom/create-terraform-resources-only-in-certain-environments-dde0f20578a5).

main.tf
```diff
@@ -0,0 +1,166 @@
...
+
+locals {
+  s3_prefix                = "scripts/sales-val"
+  s3_bucket_data_warehouse = terraform.workspace == "prod" ? "ccao-data-warehouse-us-east-1" : aws_s3_bucket.data_warehouse[0].id
+  s3_bucket_glue_assets    = terraform.workspace == "prod" ? "ccao-glue-assets-us-east-1" : aws_s3_bucket.glue_assets[0].id
+  glue_job_name            = "sales_val_flagging${terraform.workspace == "prod" ? "" : "_${terraform.workspace}"}"
+  glue_crawler_name        = "ccao-data-warehouse-sale-crawler${terraform.workspace == "prod" ? "" : "-${terraform.workspace}"}"
+}
+
+variable "iam_role_arn" {
```
jeancochrane, line 31: [Variables](https://developer.hashicorp.com/terraform/language/values/variables) are values that must either be passed in by the Terraform caller, or set with a `default`. We keep the `iam_role_arn` out of version control because it contains a reference to our AWS Account ID, which [is not technically a secret but should be used cautiously](https://docs.aws.amazon.com/accounts/latest/reference/manage-acct-identifiers.html). The IAM role itself is not managed by this Terraform config because it is shared by other Glue jobs, so it will be defined in our upcoming org-wide Terraform config.

main.tf
```diff
@@ -0,0 +1,166 @@
...
+variable "commit_sha" {
+  type        = string
+  description = <<EOF
+    "Hash of the current git commit, to use for versioning the model. "
+    "Generate with `git rev-parse --short HEAD`"
+  EOF
+  nullable    = false
+}
+
+resource "aws_s3_bucket" "glue_assets" {
```
jeancochrane, line 47: [Resource blocks](https://developer.hashicorp.com/terraform/language/resources/syntax) define the AWS entities that we want Terraform to manage (i.e. create, update, or destroy). The docs for the [Terraform AWS provider](https://registry.terraform.io/providers/hashicorp/aws/latest/docs) can be useful for deciphering what each of these resources and their attributes mean, but in general everything should have a pretty straightforward 1:1 correspondence with an entity or its attribute in AWS.

main.tf
```diff
@@ -0,0 +1,166 @@
...
+  versioning_configuration {
+    status = "Enabled"
+  }
+}
+
+resource "aws_s3_object" "sales_val_flagging" {
+  bucket = local.s3_bucket_glue_assets
+  key    = "${local.s3_prefix}/sales_val_flagging.py"
+  source = "${path.module}/glue/sales_val_flagging.py"
+  etag   = filemd5("${path.module}/glue/sales_val_flagging.py")
```
jeancochrane, line 102: Combined, these two attributes allow us to upload a new version of the script to S3 whenever its contents change. [`source`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_object#source) indicates the path to the file in the local filesystem that should be uploaded to S3, while [`etag`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_object#etag) defines a value (in this case, the md5 hash of the file) that determines whether the resource should be updated or not -- i.e. if the hash changes, the file must be uploaded.

jeancochrane:
Something weird is going on with the state here -- reverting to a draft until I figure it out!

main.tf
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  required_version = ">= 1.5.7"
+
+  backend "s3" {
+    bucket               = "ccao-terraform-state-us-east-1"
+    key                  = "terraform.tfstate"
+    region               = "us-east-1"
+    workspace_key_prefix = "model-sales-val/workspaces"
+  }
```
jeancochrane, line 16: This block defines the [backend](https://developer.hashicorp.com/terraform/language/settings/backends/configuration) that Terraform will use to manage state. The `key` attribute defines the location of the default workspace, but we don't use that workspace at all in our GitHub Actions config, so workspace state files will instead follow the naming scheme set by `workspace_key_prefix` and will be stored under `s3://ccao-terraform-state-us-east-1/workspaces/<workspace>/terraform.tfstate`.

jeancochrane:
Sorry for the confusion, this is once again ready for review 👍🏻

dfsnow (changes requested):
@jeancochrane Thanks for this excellent PR. I really appreciate the additional context and explanations around the terraform bits.

My only concern here is around how the actual output of the Glue job is created and tested. Right now, the job pulls from the production `sale.flag` table to determine which sales need a new flag. Since that table is prod and already populated, there are no _new_ sales to flag, thus the test job outputs nothing to the bucket. I suggest we fix this by creating a separate `sale.flag_test` table that is targeted and populated by PRs only. We would need to:

- Instantiate a new Athena table with the same schema as `sale.flag`, but pointing to the test bucket location
- Edit the input query in the Glue job to conditionally point to the test table during PRs, and to the prod table otherwise
- (Optionally) Run the PR Glue job once resources exist in order to populate the test table

For manual updates to the prod data, we can still use `manual_update.py` to write new runs to the prod table. Each run is a parquet file that can be easily deleted if necessary.

.github/actions/setup-terraform/action.yaml
```diff
@@ -0,0 +1,69 @@
+---
+name: Setup Terraform
+description: Install and configure Terraform and AWS for the correct environment
+inputs:
+  role-to-assume:
+    description: AWS IAM role to assume when running Terraform operations.
+    required: true
+  terraform-iam-role-arn:
+    description: AWS IAM role to use for running the Glue job being deployed.
+    required: true
```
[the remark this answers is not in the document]
jeancochrane: Good question! The role for GitHub Actions to assume (`role-to-assume`) was scoped as tight as I could get it around the specific resources that Terraform needs to create, update, or destroy for this repo. The role that Glue assumes (`terraform-iam-role-arn`) is the pre-existing role that the job is already using, and it seems less tightly scoped. I'll send you the ARNs over Teams so you can examine them yourself.

.github/workflows/cleanup-terraform.yaml
jeancochrane: Yup, that's right!

glue/sales_val_flagging.py
jeancochrane: Cool, I double-checked that the script output is correct and added a note to the PR body reminding myself that we need to recreate and recrawl this data after merging 👍🏻

main.tf
jeancochrane: Under this namespacing scheme, the fact that the top level of the prefix path corresponds to the repo name is intended to protect against that case! So for example, if we had a branch `jean-testing` in both the `data-architecture` and `model-sales-val` repos, the state paths would look like:

* `s3://ccao-terraform-state-us-east-1/model-sales-val/workspaces/jean-testing/terraform.tf`
* `s3://ccao-terraform-state-us-east-1/data-architecture/workspaces/jean-testing/terraform.tf`

The one exception to this would be the `default` workspace state file, which both projects (if they used a similar config as this one) would try to store under `s3://ccao-terraform-state-us-east-1/terraform.tfstate`. I'm assuming we'll handle that by never using `default` workspaces in cases where we have `workspace_key_prefix` set, but an alternative would be to have different state buckets for each project; that feels like overkill, but I'm open to alternative opinions here!

.github/actions/setup-terraform/action.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+    - name: Select Terraform workspace
+      run: |
+        # # GITHUB_HEAD_REF is only set on pull_request events, so if it's
+        # present, we must be in a PR context
+        if [ -n "$GITHUB_HEAD_REF" ]; then
+          echo "On pull request branch, setting terraform workspace to CI"
+          # Replace slashes with hyphens in the workspace name, since we need
+          # to use it to name S3 buckets
+          WORKSPACE="${GITHUB_HEAD_REF//\//-}"
```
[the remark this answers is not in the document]
jeancochrane: Good thinking! Fixed to handle underscores in [3852544](https://github.com/ccao-data/model-sales-val/pull/31/commits/38525445c733bae1f0f0810df26758318be453f1).

wagnerlmichael (approved):
Thanks for the intro to terraform! Looking forward to learning more once we integrate.

main.tf
dfsnow: Got it got it, thanks for explaining! Current state works great!

dfsnow: Got it got it, thanks for explaining! Current state works great!

glue/sales_val_flagging.py
```diff
@@ -562,7 +553,7 @@ def write_to_table(df, table_name, s3_warehouse_bucket_path, run_id):
         WHERE condo.class IN ('297', '299', '399')
         AND NOT condo.is_parking_space
         AND NOT condo.is_common_area
-        AND condo.is_question_garage_unit IS NULL
+        AND NOT condo.is_question_garage_unit
```
jeancochrane, line 556: Michael helped figure out that we need to make this change in order to return any condos: https://github.com/ccao-data/data-architecture/pull/116#issuecomment-1706939507

glue/sales_val_flagging.py
```diff
@@ -572,7 +563,7 @@ def write_to_table(df, table_name, s3_warehouse_bucket_path, run_id):
         INNER JOIN default.vw_pin_sale sale
             ON sale.pin = data.pin
             AND sale.year = data.year
-        LEFT JOIN sale.flag flag
+        LEFT JOIN {args["sale_flag_table"]} flag
```
jeancochrane, line 566: New as of last review cycle, we are now parameterizing the tablename based on the environment! Staging environments will pull from a staging version of `sale.flag` created for the PR.

main.tf
```diff
@@ -0,0 +1,214 @@
...
+  etag   = filemd5("${path.module}/glue/flagging_script_glue/flagging.py")
+}
+
+resource "aws_s3_object" "sale_flag_metadata" {
+  count  = terraform.workspace == "prod" ? 0 : 1
+  bucket = local.s3_bucket_data_warehouse
+  key    = "sale/flag/sale_flag_schema.parquet"
+  source = "${path.module}/glue/schema/sale_flag_schema.parquet"
+  etag   = filemd5("${path.module}/glue/schema/sale_flag_schema.parquet")
+}
```
jeancochrane, line 148: New as of last review cycle, this bucket stores query output for the staging Athena database.

main.tf
```diff
@@ -0,0 +1,214 @@
...
+  }
+}
+
+resource "aws_athena_database" "sale_test" {
+  count         = terraform.workspace == "prod" ? 0 : 1
+  name          = local.athena_database_name
+  comment       = "Test sale database for the ${terraform.workspace} branch"
+  bucket        = aws_s3_bucket.athena_results[0].id
+  force_destroy = true
+}
```
jeancochrane, line 192: New as of last review cycle, this staging database lets us keep our testing data isolated from production data.

main.tf
```diff
@@ -0,0 +1,214 @@
...
+  s3_prefix                = "scripts/sales-val"
+  s3_bucket_data_warehouse = terraform.workspace == "prod" ? "ccao-data-warehouse-us-east-1" : aws_s3_bucket.data_warehouse[0].id
+  s3_bucket_glue_assets    = terraform.workspace == "prod" ? "ccao-glue-assets-us-east-1" : aws_s3_bucket.glue_assets[0].id
+  glue_job_name            = "sales_val_flagging${terraform.workspace == "prod" ? "" : "_${terraform.workspace}"}"
+  glue_crawler_name        = "ccao-data-warehouse-sale-crawler${terraform.workspace == "prod" ? "" : "-${terraform.workspace}"}"
+  # Athena databases cannot have hyphens, so replace them with underscores
+  # (Note that this is not always true -- notably, dbt-athena is able to
+  # create Athena tables with hyphens -- but it's a rule that Terraform
+  # enforces, so we follow it here)
+  athena_database_name = terraform.workspace == "prod" ? "sale" : "ci_model_sales_val_${replace(terraform.workspace, "-", "_")}_sale"
```
jeancochrane, line 33: New as of last review cycle, this database helps keep staging data isolated on branches. The mismatch with our dbt staging database pattern is a bit of a bummer 😕

jeancochrane:
@dfsnow @wagnerlmichael After much debugging, we have a successful staging run, so this is ready for another look! See[ here](https://us-east-1.console.aws.amazon.com/gluestudio/home?region=us-east-1#/editor/job/sales_val_flagging_jeancochrane-terraform-spike/runs) for the successful run, and try this query to check staging table state after the test run:

```sql
SELECT count(*)
FROM ci_model_sales_val_jeancochrane_terraform_spike_sale.flag flag

> 32729
```

wagnerlmichael (commented):
On my end, I just looked at the output of the `sale.flag`. Although the flagging is not correct (I'm working to fix that in two different outstanding branches), it seems to be doing exactly what it is supposed to be doing!

One quick note that I'm not sure about - it seems like we don't have any ptax flags for 2023. I noticed because Jean chose only 2023 data as the subset. No ptax flags appeared in the `sv_outlier_type`.

Are those put in once a year in a batch upload or something? I checked the `vw_pin_sale` and they don't seem to be there either @dfsnow

glue/schema/generate_sale_flag_schema.py
dfsnow, line 1: **suggestion (non-blocking):** This totally works, but is it possible to run a `CREATE TABLE` statement using `aws athena start-query-execution $QUERY`? It might be simpler and eliminate the Python dependency. I'm guessing this is the permissions complication though.

main.tf
dfsnow: **suggestion:** You're gonna hate me for this, but can we prefix the ephemeral buckets with `ci-` or `dev-`, as we did for the Athena resources? It keeps them separately sorted in the UI and also helps differentiate their purpose.

main.tf
```diff
@@ -95,6 +100,31 @@ resource "aws_s3_bucket_versioning" "data_warehouse" {
...
+
+resource "aws_s3_bucket_versioning" "athena_results" {
+  count  = terraform.workspace == "prod" ? 0 : 1
+  bucket = aws_s3_bucket.athena_results[0].id
+
+  versioning_configuration {
+    status = "Enabled"
+  }
+}
+
```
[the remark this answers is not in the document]
jeancochrane: Cool, I just wasn't sure whether it was OK to use the prod bucket! I switched to use it in efd94bab.

main.tf
jeancochrane: No problem, done in efd94bab!

glue/schema/generate_sale_flag_schema.py
jeancochrane: I thought it wasn't possible, but after a little bit more research I figured out that we can use the [`aws_glue_catalog_table`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/glue_catalog_table) resource to wrap this operation in a way that Terraform will understand! Implemented in 939c93a5. Good call, this simplifies things a lot 👍🏻

README.md
```diff
@@ -121,45 +120,27 @@ The current implementation uses a 12 month rolling window. This means that for a
...
-    - Removes the previously hashed flagging file.
-4. The updated hash file should reflect in both the S3 bucket and this repository.
-5. Push the changes to the master branch.
-
+    - Make a new git branch for your changes.
+    - Edit the files as necessary.
+    - Open a pull request for your changes against the `main` branch. A GitHub Actions workflow called `deploy-terraform` will deploy a staging version of your job named `sales_val_flagging_<your_branch_name>` that you can run to test your changes.
+      - If you would like to test your job against a subset of the production data, copy your data subset from the production job bucket to the bucket created by Terraform for your job (or leave the new bucket empty to simulate running the job when no flags exist). Then, run the crawler created by Terraform for your PR in order to populate the staging version of the `sale.flag` database that your staging job uses. If you're having trouble finding your staging bucket, job, or crawler, check the GitHub Actions output for the first successful run of your PR and look for the Terraform output displaying the IDs of these resources.
```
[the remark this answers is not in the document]
jeancochrane: Yup, as of 939c93a5 we won't need to run the crawler anymore! I added a line to the README text in 6117f4d7 to make this clear, but I ended up leaving this block mostly in place as an explanation of how a developer could simulate production data by copying it over and running the crawler.

dfsnow (approved):
Nice simplification @jeancochrane! Let's merge this today and have @wagnerlmichael do a test run using a new PR.

PR #34, opened 2023-10-10
Add fix for - #33 square foot flags not outputting

### LINKED ISSUE #33: Square foot flags not outputting
In the `sv_outlier_type` column, there are no entries for sqft outlier types. We expect to see these flags from res sales.

wagnerlmichael:
The prior refactor changed the order of conditions in `outlier_type()`, which was overwriting the sqft flags. I reverted to the old if/else logic, and it functions as expected. I tried for a while to refactor to make the function cleaner, but the best I could get is about 10 lines less of code that was overall less readable. I think we should just leave the longer conditional blocks as is.

Closes #33


dfsnow (approved):
@wagnerlmichael You have more visibility on the correctness of these conditionals than I do, so I'm going to trust that this PR is correct. The code itself looks fine.

PR #36, opened 2023-10-10
35 remove old hashed py files

### LINKED ISSUE #35: Remove old hashed py files
We need to remove old hash `.py` files, as they are an artifact of pre-terraform infrastructure. We also need to update the recent changes from [this PR](https://github.com/ccao-data/model-sales-val/pull/34), so that they are reflected in `glue/flagging_script_glue/flagging.py`, which is built into the terraform architecture.

wagnerlmichael:
Cleaned out hashed files and updated the `glue/flagging_script_glue/flagging.py` file with recent changes from PRs that were already reflected in the local flagging file inside `src/`.

Closes #35


jeancochrane (approved):
Looks good, thanks for the quick fix!

PR #39, opened 2023-10-13
38 Improve string matching

### LINKED ISSUE #38: Improve string matching
There is a significant amount of regex work in this program that helps to flag sales as outliers in the following categories:
- Family Sale
- Non-person Sale

Identifying these sales can help us expose non-arms length sales, such as when an institutional buyer is involved, or when a family sale happens. We need to dig into the code and improve these classifiers.

wagnerlmichael:
Opening this draft PR now to test terraform integration

Edit: Made a number of small fixes to string matching and classification of outliers. Let me know what you think.

Closes #38

### DIFF
--- a/glue/flagging_script_glue/flagging.py
+++ b/glue/flagging_script_glue/flagging.py
@@ -728,7 +728,11 @@ def outlier_type(df: pd.DataFrame, condos: bool) -> pd.DataFrame:
             & (df["sv_pricing"].str.contains("High")),
             (df["sv_name_match"] != "No match")
             & (df["sv_pricing"].str.contains("High")),
-            (df["sv_transaction_type"] == "legal_entity-legal_entity")
+            (
+                df[["sv_buyer_category", "sv_seller_category"]]
+                .eq("legal_entity")
+                .any(axis=1)
+            )
             & (df["sv_pricing"].str.contains("High")),
             # (df["sv_anomaly"] == "Outlier") & (df["sv_pricing"].str.contains("High")),
             (df["sv_pricing"].str.contains("High price swing")),
@@ -737,7 +741,11 @@ def outlier_type(df: pd.DataFrame, condos: bool) -> pd.DataFrame:
             & (df["sv_pricing"].str.contains("Low")),
             (df["sv_name_match"] != "No match")
             & (df["sv_pricing"].str.contains("Low")),
-            (df["sv_transaction_type"] == "legal_entity-legal_entity")
+            (
+                df[["sv_buyer_category", "sv_seller_category"]]
+                .eq("legal_entity")
+                .any(axis=1)
+            )
             & (df["sv_pricing"].str.contains("Low")),
             # (df["sv_anomaly"] == "Outlier") & (df["sv_pricing"].str.contains("Low")),
             (df["sv_pricing"].str.contains("Low price swing")),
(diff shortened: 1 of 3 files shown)


glue/flagging_script_glue/flagging.py
```diff
@@ -728,7 +728,11 @@ def outlier_type(df: pd.DataFrame, condos: bool) -> pd.DataFrame:
             & (df["sv_pricing"].str.contains("High")),
             (df["sv_name_match"] != "No match")
             & (df["sv_pricing"].str.contains("High")),
-            (df["sv_transaction_type"] == "legal_entity-legal_entity")
+            (
+                df[["sv_buyer_category", "sv_seller_category"]]
+                .eq("legal_entity")
+                .any(axis=1)
+            )
```
wagnerlmichael, line 735: We changed this which allows for `non-person` flag to only require one `legal-entity` between the buyer and seller.

glue/flagging_script_glue/flagging.py
```diff
@@ -893,7 +909,7 @@ def outlier_flag(df: pd.DataFrame) -> pd.DataFrame:
     r"|associates|consultants|international|acquisitions|credit|design"
     r"|homeownership|solutions|home|diversified|assets|family|land|"
     r"revocable|services|rehabbing|living|county of cook|fannie mae|"
-    r"land|veteran|mortgage|savings|lp$"
+    r"land|veteran|mortgage|savings|lp$|federal natl"
```
wagnerlmichael, line 912: The regex wasn't picking this up - an institutional entity.

glue/flagging_script_glue/flagging.py
```diff
@@ -938,6 +954,9 @@ def get_id(row: pd.Series, col: str) -> str:
     if any(x in words for x in ["vt investment corpor", "v t investment corp"]):
         return "vt investment corporation"
 
+    if any(x in words for x in ["national residential nomi"]):
+        return "national residential nominee services"
+
```
wagnerlmichael, line 959: This entity wasn't being flagged as a `legal-entity` either.

glue/flagging_script_glue/flagging.py
```diff
@@ -983,6 +1002,7 @@ def get_id(row: pd.Series, col: str) -> str:
             "wells fargo bank n a",
             "wells fargo bank nationa",
             "wells fargo bank n a a",
+            "wells fargo bk",
```
wagnerlmichael, line 1005: We were missing this form.

glue/flagging_script_glue/flagging.py
```diff
@@ -1036,6 +1056,8 @@ def get_id(row: pd.Series, col: str) -> str:
             "u s bank n a",
             "us bank national associat",
             "u s bank trust national",
+            "us bk",
+            "u s bk",
```
wagnerlmichael, line 1060: We missed these as well

glue/flagging_script_glue/flagging.py
```diff
@@ -1257,7 +1279,14 @@ def create_name_match(row: pd.Series) -> str:
...
     """
-    if row["sv_buyer_id"] == row["sv_seller_id"] and row["sv_buyer_id"] != "Empty Name":
+    if (
+        row["sv_buyer_id"] == row["sv_seller_id"]
+        and row["sv_buyer_id"] != "Empty Name"
+        # Prevents the same legal entity as counting as a family name match
+        and row["sv_transaction_type"] != "legal_entity-legal_entity"
+        # Boots out matches on a single last initial
+        and len(row["sv_buyer_id"]) > 1
+    ):
```
wagnerlmichael, line 1289: There are two fixes here:

- First, if a legal entity was both a buyer and a seller with the exact string name, it would be classified as a `Family sale`. This line of code fixes that: `and row["sv_transaction_type"] != "legal_entity-legal_entity"`
- Second, there were multiple instances where the last token picked up by the regex was a single initial which resulted in a family sale name match. We fix that here: `and len(row["sv_buyer_id"]) > 1`

glue/flagging_script_glue/flagging.py
dfsnow: **question:** To clarify, this will now generate a `non-person` flag when one side of the transaction is a legal entity AND the sale falls outside the `2,3` SD bounds (or there's another type of statistical flag)?

glue/flagging_script_glue/flagging.py
dfsnow: **praise:** Nice! Good catches.

dfsnow (approved):
Looks good @wagnerlmichael. Nice work.

glue/flagging_script_glue/flagging.py
wagnerlmichael: That's correct, both conditions need to be satisfied.

PR #42, opened 2023-10-17
Update date floor

### LINKED ISSUE #41: Update glue job date floor for production
We need to set the floor of the data ingest for the glue job for 2014-01-01, as we want the glue job to flag any unflagged sales from 2014 onward.

wagnerlmichael:
Switching the date floor to get ready for production and do a final test.

Closes #41


jeancochrane (approved):
Looks like the build is failing due to a problem with our Terraform config; I opened an issue to fix it (https://github.com/ccao-data/model-sales-val/issues/43) but in the meantime I think I or Dan should be able to override the status check and deploy anyway!

This change looks good to me, but I haven't tested it and I'm assuming you've tested it already!

PR #44, opened 2023-10-17
Improved String Detection

Damonamajor:
**I asked Michael about this, but I don't really know the full workflow for sales-val, and while these are minimal changes, I couldn't check if it breaks the larger code in unexpected ways.**

I grouped the seller and buyer strings by the most common full words and then manually went through and identified words that likely signified institutional sales. I got down to around 30 occurrences by both buyer/seller. They are listed below in a framework of how I understand the pull request merge. \b \b around the words references full word requirements. 

**No questions:**
hospital
southport
propert
rehab
neighborhood
advantage
chicago
\bbk\b - short for bank
\bhud\b
department
united states
\busa\b
hsbc
midwest
residential
american
tcf
advantage
mtg - short for mortgage
\bfed\b
nationstar
commercial
health
condominum
nationa 
association

**Involves Discussion Before Merge**
trustee: See comment at bottom
successor - similar logic to trustee
real e - real estate doesn't catch everything and real can be a last name
cook c - cook by itself can be a last name and cook county is not always written out
baptist church
apostolic church
lutheran church - church can be a last name

**Doubtful Merge for the Moment**
amp - I think this stands for ampersand but am not sure
CO is another key word, which is often reference with CO-Trustee, but can also be short for corporation

**Modified exiting code:**
Put land in /b /b because last names such as Bjorkland/kneeland were reclassified
Put Home in /b /b because first names such as Homer were reclassified
Put Sale in /b /b because first names such as Saleem and last names like Salem were reclassified
Put Loan in /b /b because names such as Sloan were being reclassified
Put living in /b /b because livingston was being reclassified

**Unsolved things to note.**
b/the /b as coded is mostly right but does catch things like  The Christopher John Hall
Family is another key word, which is removing a lot of trusts. This isn't a huge issue, but it does mean that the Snow Family Trust will be recoded, but the D. Snow Trust will not be recoded.
I could imagine CORP being part of names, but couldn't find any. There are also many misspellings of corporation, so leaving as is seems the best option. 
Jack Buyers and Robert Buyers are names. These are 2 out of 22 observations for buyers.
Homes does not seem to be a last name.

**Tustee Logic**
Trustee is the largest question mark. There are approximately 7,000 unique observations in the database regarding "people sales"(only larger categories are Michael, John, James, and Robert). These could represent a particular type of sale (to kids or family members). Because this hasn't been discussed, I'm only going to make a note of it, although Mansuetto mentioned it in their logic.


wagnerlmichael (approved):
This looks great to me, thanks for catching all of these and adding some quality-of-life improvements. As for the trustee issue -  I don't have enough domain knowledge to have a strong opinion so I will differ to @dfsnow. I believe the current strategy it to remove the trustee tag, grab the last name, and treat it as a person sale.

For the phrases you marked as to discuss, I think being more strict is probably better right now. I'm thinking the entire church string for those specific churches are good to keep in the `entity_keywords`.

dfsnow:
@Damonamajor Can you resolve the merge conflict here and fixup the main body of the issue to use proper markdown formatting?

Damonamajor:
> @Damonamajor Can you resolve the merge conflict here and fixup the main body of the issue to use proper markdown formatting?

@dfsnow - Done

dfsnow:
@Damonamajor I agree with @wagnerlmichael here. Let's add strings to catch specific, non-last names churches. We can ignore the trustee stuff for the time being. After that, this is good to go.

Damonamajor:
Just as a final note, I added a few more churches to the matching. And, before I merge, I pasted the same changes in the other location that the string detection is used - manual_flagging/src/flagging_rolling.py.

PR #47, opened 2023-10-18
46 update run id to reflect the ccao package standard

### LINKED ISSUE #46: Update run-id to reflect the ccao package standard
Currently we are assigning run-id using a random word generator - we need to change this so that we use the standard run-id generator that we use in the ccao package.

wagnerlmichael:
Switched out the run_id methodology per @ccao-jardine's request. I copy and pasted the data from the ccao package.

Closes #46


dfsnow (approved):
Works for me. Nice job @wagnerlmichael.

dfsnow (commented):
@wagnerlmichael This works fine, but there might be a way to keep the words synchronized between this and the CCAO package. It might be worth looking into git submodules.

PR #51, opened 2023-10-18
Fixed Reg-ex Error

Damonamajor:
Fixed || in previous Regex Push.


wagnerlmichael (approved):
Looks good, good catch.

PR #52, opened 2023-10-18
Remove local flagging script, use glue script since they are identical

### LINKED ISSUE #50: De-duplicate flagging script
Right now we have a separate script for S3/glue and for the local scripts. The files are identical, we need to delete the script in `src/` folder and pull in the flagging script from glue into the local script. This is better as we will have a single source of truth for the flagging model.

wagnerlmichael:
Deleted local flagging scripts, and now the flagging script in the glue directory is used in our local scripts.

Closes #50


dfsnow (commented):
This makes sense to me as long as you've confirmed it works via a Terraform test setup. @jeancochrane any quick thoughts?

jeancochrane (approved):
I'm not totally sure why it was like this before, but the change seems fine to me!

dfsnow:
> I'm not totally sure why it was like this before, but the change seems fine to me!

IIRC it was an artifact of the fact that we had to hash the flagging script and thus couldn't reference it via `import`. Or something like that 🤷 .

wagnerlmichael:
Yep that's correct, also because I had to have different code for each script because of the pandas mismatch.

PR #54, opened 2023-10-19
Remove flag table integration from recurring script

wagnerlmichael:
I tested this with a date floor of `2023-01-01`. It successfully writes the data, and then when we run it again afterwards, it prints `"WARNING: No new sales to flag"` from line 783. Final step is for me to run the production data and inspect the results after merging.

### DIFF
--- a/glue/sales_val_flagging.py
+++ b/glue/sales_val_flagging.py
@@ -875,12 +875,22 @@ if __name__ == "__main__":
             start_date=args["time_frame_start"],
             manual_update=False,
         )
-
+        """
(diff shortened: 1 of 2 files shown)


dfsnow (approved):
Nice @wagnerlmichael. This looks good to me. We'll have to edit this further as Billy gets traction on splitting the sales view (https://github.com/ccao-data/data-architecture/issues/149) and Jean fixes up the terraform testing bit here (#55).

jeancochrane (approved):
Looks good to me as well!

Issue #53, opened 2023-10-19
Update recurring scripts to use updated view


wagnerlmichael:
Since we have [updated the sales view](https://github.com/ccao-data/data-architecture/commit/502ca193eea7e3291c754f9637fe7da49a3c316d), we can now edit `glue/sales_val_flagging.py` so that we can utilize the sales val columns from within the view, rather than making a bunch of joins on the `sale.flag` table.

Issue #55, opened 2023-10-20
Update Terraform configuration to deploy staging `vw_pin_sale` view


jeancochrane:
In working on https://github.com/ccao-data/model-sales-val/issues/53, @wagnerlmichael discovered that the testing version of script won't work as written because it pulls from the production `vw_pin_sale` view, which itself pulls from the production `sale.flag` table instead of the staging version of that table. In order to support removing the joins on `sale.flag` from the script, we'll need to update our Terraform config to deploy a staging version of `vw_pin_sale` that will pull data from the staging `sale.flag` table.

jeancochrane:
A hiccup with this effort that I should have thought of sooner: Unlike `sale.flag`, which is generated by code in this repo, the `vw_pin_sale` view is actually created by [code that lives in in `data-architecture`](https://github.com/ccao-data/data-architecture/blob/master/aws-athena/views/default-vw_pin_sale.sql). This means that if we wanted to deploy a staging version of it from the context of `model-sales-val`, we would need to either duplicate the view query in `model-sales-val` or somehow reference the version of it stored in the`data-architecture` repo, neither of which is a particularly appealing choice to me.

Two options off the top of my head:

1. Revert [this section of #54](https://github.com/ccao-data/model-sales-val/pull/54/files#diff-1416936c88079b0895b64ce5aa2ca9bb94dc585671b364426100fc401f0dce5aR724-R746) and maintain a duplicative join between `vw_pin_sale` to `sale.flag` in the context of the Glue script
    * Pro: Dead simple and easy to maintain
    * Con: Less efficient and potentially more confusing query due to the duplication
2. Update the deployment workflow and Terraform config so that on staging deployments we download the `vw_pin_sale` query from the main branch of `data-architecture` and use it to populate [an `aws_glue_catalog_table` resource](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/glue_catalog_table)
    * Pro: Eliminates duplication in the Glue job query
    * Con: Fragile for a number of reasons:
        * Must be updated if the file structure of `data-architecture` ever changes
        * Makes local testing of the Terraform setup more complex
        * Still requires some duplication with `data-architecture`, since I confirmed experimentally that this Terraform resource[ requires each column in the view be specified](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/glue_catalog_table#columns)

Option 1 is my preference but I'm curious if @wagnerlmichael or @dfsnow have other ideas!

[...]

PR #57, opened 2023-10-23
Revert query changes to use sale.flag

### LINKED ISSUE #56: Revert glue job to join with sale.flag
In #53 we changed the data ingest and newly flagged sales to use the updated sales view, rather than making joins with the `sale.flag` table. This significantly increases the complexity in the terraform testing setup, so we need to revert this change for now. Functionally, nothing will change, but the ingest will be less efficient.

wagnerlmichael:
Reverted changes from #53.

Closes #56.

### DIFF
--- a/glue/sales_val_flagging.py
+++ b/glue/sales_val_flagging.py
@@ -763,11 +763,6 @@ if __name__ == "__main__":
     )
     """
 
-    SQL_QUERY_SALES_VAL = f"""
-    SELECT *
-    FROM {args["sale_flag_table"]}
-    """
-
     # -------------------------------------------------------------------------
     # Execute queries and return as pandas df
     # -------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


dfsnow:
Related to #55.

dfsnow (approved):
Looks good to me @wagnerlmichael. Thanks for the quick fix.

glue/sales_val_flagging.py
```diff
@@ -759,6 +763,11 @@ def write_to_table(df, table_name, s3_warehouse_bucket_path, run_id):
     )
     """
 
+    SQL_QUERY_SALES_VAL = f"""
+    SELECT *
+    FROM {args["sale_flag_table"]}
+    """
```
jeancochrane, line 769: **[Question, non-blocking]** I think it's not necessary to revert the section of #53 that removed this code block and altered the subsequent code to avoid a duplicate query, right? It seems to me that it's unrelated to the `sale.flag` join but I could be wrong 🤔

jeancochrane (approved):
One concern that we're reverting unrelated code that we want to keep, but other than that this looks good to me!

jeancochrane (approved):
Good to go!

PR #59, opened 2023-10-24
Switch is columns from 01 to boolean

### LINKED ISSUE #58: Switch `is_` columns from 0/1 to boolean.
We need to switch `sv_is_outlier`, `sv_is_ptax_outlier`, and `sv_is_heuristic_outlier` to boolean columns instead of 0/1.

wagnerlmichael:
Switched columns from `1/0` to `True/False`.

Closes #58

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,10 +64,11 @@ erDiagram
     flag {
         string meta_sale_document_num PK
         date rolling_window
-        bigint sv_is_outlier
-        bigint sv_is_ptax_outlier
-        bigint sv_is_heuristic_outlier
-        string sv_outlier_type
+        boolean sv_is_outlier
+        boolean sv_is_ptax_outlier
+        boolean sv_is_heuristic_outlier
+        boolean ptax_flag_original
+        boolean sv_outlier_type
         string run_id FK
         bigint version PK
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -64,10 +64,11 @@ erDiagram
...
         date rolling_window
-        bigint sv_is_outlier
-        bigint sv_is_ptax_outlier
-        bigint sv_is_heuristic_outlier
-        string sv_outlier_type
+        boolean sv_is_outlier
+        boolean sv_is_ptax_outlier
+        boolean sv_is_heuristic_outlier
+        boolean ptax_flag_original
+        boolean sv_outlier_type
```
dfsnow, line 71: **nitpick:** I would put these in the order they appear in the database. In Athena they appear in a different order, but it could be because of an old crawl.

dfsnow (approved):
@wagnerlmichael This looks good to me. You'll likely have to do another full run + a recrawl. Let's do that later today once all the sales view stuff is also merged.

README.md
wagnerlmichael: The order of those columns are selected downstream when we subset by columns.

wagnerlmichael: Fixed

PR #61, opened 2023-10-25
Modified string matching for lawyer and Delaware

Damonamajor:
The switch from law to /blaw/b got rid of many erroneous flags, but also removed accurate flags. These mostly related to the words lawyer and Delaware. 

Delaware is often truncated, so shortened versions of the word are used to optimize the matching.

### DIFF
--- a/glue/flagging_script_glue/flagging.py
+++ b/glue/flagging_script_glue/flagging.py
@@ -916,7 +916,7 @@ entity_keywords = (
     r"|apostolic church|lutheran church|catholic church|\bfed\b|nationstar"
     r"|advantage|commercial|health|condominium|nationa|association|homeowner"
     r"|christ church|christian church|baptist church|community church"
-    r"|church of c"
+    r"|church of c|\bdelaw\b|lawyer|delawar"
 )
 
 


glue/flagging_script_glue/flagging.py
```diff
@@ -916,7 +916,7 @@ def outlier_flag(df: pd.DataFrame) -> pd.DataFrame:
     r"|apostolic church|lutheran church|catholic church|\bfed\b|nationstar"
     r"|advantage|commercial|health|condominium|nationa|association|homeowner"
     r"|christ church|christian church|baptist church|community church"
-    r"|church of c"
+    r"|church of c|\bdelaw\b|lawyer|delawar"
```
wagnerlmichael, line 919: These changes look good to me. It seems like the DBT build is failing, and I do not know why. This might be of interest to @jeancochrane.

jeancochrane: Looks like the uppercase letters in the branch name for this PR violate Athena database naming rules:

```
╷
│ Error: invalid value for name (must be lowercase letters, numbers, or underscore ('_'))
│ 
│   with aws_athena_database.sale,
│   on main.tf line [16](https://github.com/ccao-data/model-sales-val/actions/runs/6642150013/job/18046234338?pr=61#step:4:17)9, in resource "aws_athena_database" "sale":
│  169:   name          = local.athena_database_name
│ 
╵
```

I can put up a quick PR to fix this!

PR #62, opened 2023-10-25
Modified string matching for `lawyer` and `delaware`

Damonamajor:
The switch from law to /blaw/b got rid of many erroneous flags, but also removed accurate flags. These mostly related to the words lawyer and Delaware.

Delaware is often truncated, so shortened versions of the word are used to optimize the matching.


dfsnow (approved):
This looks good to me, but I'll defer to @wagnerlmichael re: the final utility of these changes.

wagnerlmichael (approved):
This looks good to me! It should make the string matching more accurate.

PR #64, opened 2023-10-25
Switch Glue job type to Spark instead of Python shell

jeancochrane:
This PR is primarily authored by @wagnerlmichael. [Michael's description below](https://github.com/ccao-data/model-sales-val/pull/64#issuecomment-1781375188):

> This PR adjusts the glue script to a spark job in order to take advantage of more compute resources. In the old python shell job, we would get an out of memory error if the ingest took in too much data.
>
> I've tested both glue and local - both run without issue and the proportion of flags are almost exactly the same. The work involved mostly troubleshooting package changes since the new spark glue job does not support some of the package versions we had been using. I standardized the package changes in local and in glue.

### DIFF
--- a/glue/flagging_script_glue/flagging.py
+++ b/glue/flagging_script_glue/flagging.py
@@ -543,15 +543,16 @@ def get_sale_counts(dups: pd.DataFrame) -> pd.DataFrame:
     Calculates how many times transactions occured for a gieven property.
     Helper for dup_stats()
     Inputs:
-        df (pd.DataFrame): pandsa dataframe4
+        df (pd.DataFrame): pandas dataframe
     """
     v_counts = (
         dups.pin.value_counts()
         .reset_index()
-        .rename(columns={"count": "sv_sale_dup_counts"})
+        .rename(columns={"index": "pin", "pin": "sv_sale_dup_counts"})
     )
 
-    dups = pd.merge(dups, v_counts)
+    # Explicitly specify the merging columns
+    dups = pd.merge(dups, v_counts, on="pin")
 
     return dups
 
--- a/glue/sales_val_flagging.py
+++ b/glue/sales_val_flagging.py
@@ -786,7 +786,11 @@ if __name__ == "__main__":
     if filtered_df.sv_outlier_type.isna().sum() == 0:
         print("WARNING: No new sales to flag")
     else:
+        print(df)
+        print(df.dtypes)
+        print(metadata)
         df = df.astype({col[0]: sql_type_to_pd_type(col[1]) for col in metadata})
+        print(df.dtypes)
         df["ptax_flag_original"].fillna(False, inplace=True)
 
         # Separate res and condo sales based on the indicator column
--- a/glue/sales_val_flagging.py
+++ b/glue/sales_val_flagging.py
@@ -786,9 +786,7 @@ if __name__ == "__main__":
     if filtered_df.sv_outlier_type.isna().sum() == 0:
         print("WARNING: No new sales to flag")
     else:
-        print(df)
-        print(df.dtypes)
-        print(metadata)
+        # Make sure None types aren't utilized in type conversion
         conversion_dict = {
             col[0]: sql_type_to_pd_type(col[1])
             for col in metadata
@@ -796,7 +794,6 @@ if __name__ == "__main__":
         }
         df = df.astype(conversion_dict)
 
-        print(df.dtypes)
         df["ptax_flag_original"].fillna(False, inplace=True)
 
         # Separate res and condo sales based on the indicator column
--- a/glue/sales_val_flagging.py
+++ b/glue/sales_val_flagging.py
@@ -789,7 +789,13 @@ if __name__ == "__main__":
         print(df)
         print(df.dtypes)
         print(metadata)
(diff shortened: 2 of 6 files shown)


glue/flagging_script_glue/flagging.py
```diff
@@ -543,15 +543,16 @@ def get_sale_counts(dups: pd.DataFrame) -> pd.DataFrame:
...
     v_counts = (
         dups.pin.value_counts()
         .reset_index()
-        .rename(columns={"count": "sv_sale_dup_counts"})
+        .rename(columns={"index": "pin", "pin": "sv_sale_dup_counts"})
     )
 
-    dups = pd.merge(dups, v_counts)
+    # Explicitly specify the merging columns
+    dups = pd.merge(dups, v_counts, on="pin")
```
wagnerlmichael, line 555: We had to change this code because we shifted the pandas version down.

glue/sales_val_flagging.py
```diff
@@ -786,7 +786,14 @@ def write_to_table(df, table_name, s3_warehouse_bucket_path, run_id):
...
     else:
-        df = df.astype({col[0]: sql_type_to_pd_type(col[1]) for col in metadata})
+        # Make sure None types aren't utilized in type conversion
+        conversion_dict = {
+            col[0]: sql_type_to_pd_type(col[1])
+            for col in metadata
+            if sql_type_to_pd_type(col[1]) is not None
+        }
+        df = df.astype(conversion_dict)
+
```
wagnerlmichael, line 796: This function broke because of a meta data change from athena. I'm assuming from awswragler. This stops the conversion of dtypes to `Null` which threw an error.

main.tf
```diff
@@ -156,7 +156,7 @@ resource "aws_glue_job" "sales_val_flagging" {
     "--rolling_window_num"        = 12
     "--time_frame_start"          = "2014-01-01"
     "--dev_bounds"                = "2,3"
-    "--additional-python-modules" = "boto3==1.28.12,pandas==2.0.2"
+    "--additional-python-modules" = "boto3==1.28.12,pandas==1.3.5,awswrangler==2.20.1,pyathena==2.25.2"
```
wagnerlmichael, line 159: These are the latest available packages available in the new spark configuration.

manual_flagging/requirements.txt
```diff
@@ -9,12 +9,12 @@ idna==3.4
...
-numpy==1.24.3
+numpy==1.23.4
 packaging==23.1
-pandas==2.0.2
+pandas==1.3.5
 pluggy==1.2.0
-pyarrow==12.0.1
-pyathena==3.0.6
+pyarrow==10.0.0
+pyathena==2.25.2
```
wagnerlmichael, line 17: I made the corresponding package changes to the venv in local so we can maintain a single source of truth for the flagging script, and so we can have a more consistent environment between local and glue. Changes were made to `pyarrow` and `numpy` to be compatible with the earlier package versions from the spark glue job.

wagnerlmichael:
This PR adjusts the glue script to a spark job in order to take advantage of more compute resources. In the old python shell job, we would get an out of memory error if the ingest took in too much data.

I've tested both glue and local - both run without issue and the proportion of flags are almost exactly the same. The work involved mostly troubleshooting package changes since the new spark glue job does not support some of the package versions we had been using. I standardized the package changes in local and in glue.

glue/sales_val_flagging.py
jeancochrane: **[Question, non-blocking]** What are the cases where we were finding the SQL type code in the cursor description for a column to be null?

jeancochrane (commented):
Looks good to me, although I can't approve my own PR! My only concern is that I don't totally understand what changed such that we're getting null values for SQL type codes in our cursor descriptions, but if you feel confident you know what happened there I'm fine moving forward with this as-is.

glue/sales_val_flagging.py
wagnerlmichael: My best guess is that the method functionality of `as.type()` on `pd.DataFrame` objects changed between pandas versions. Prior to the package change, I did not need to specify the `if sql_type_to_pd_type(col[1]) is not None`, it was skipping over those cases. Now, the dict comprehension seems to actually attempt to make a dtype change for each column.

The only change in the `sql_type_to_pd_type()` function is the following:

```
    if sql_type in ["decimal"]:
        return "float64"
```

So for non "decimal" types, the dict comprehension still attempts to assign a change, which is null because the dtype of the column doesn't exist in the `sql_type_to_pd_type()` function. This is my best go at understanding the error @jeancochrane

dfsnow: If you're confident that this results in no output changes then I'm satisfied with this solution @wagnerlmichael, even if it's a bit mysterious.

dfsnow (approved):
Looks good to me @wagnerlmichael and @jeancochrane. Note that I think using the Spark runners is more expensive (more DPUs), so we might want to be ever so slightly more judicious about running the full job.

wagnerlmichael:
It is ~40 cents per DPU, and a full run 2014-current uses 4 DPUS and takes 2.5 hours. Roughly $4 per run. This would be the price if we had to ingest *all* of the data every time. I'm assuming the average monthly run will be much cheaper.

PR #69, opened 2023-10-27
Edit data ingest based on vw_pine_sale updates

### LINKED ISSUE #68: Update ingest based off of sales view change
We need to update the ingest for sales val as the [underlying view](https://github.com/ccao-data/data-architecture/pull/176) from the ingest has changed.

wagnerlmichael:
Made changes to queries to reflect updated vw_pin_sale view. Tested ingest from 2014-current and the same data is ingested

Closes #68.


dfsnow (approved):
Nice! Thanks for the quick turnaround

PR #71, opened 2023-11-01
Add log10 price transformation before outlier detection

### LINKED ISSUE #65: Test log outlier detection for price columns
@dfsnow caught this issue in our flagging script. When looking to see if a price outlier is outside of our defined standard deviation bounds, we are currently normalizing the data. We want to test what the outlier flag distribution looks like when we log-transform the data *prior* to the z normalization. We want to test both log10 and ln strategies and contrast them with the existing technique. We will test all three of these strategies with [2, 2] standard deviation bounds and with [2, 3] standard deviation bounds.

wagnerlmichael:
Fix the methodology to include log 10 transformation before z normalization for flagging. More details in #65.

Closes #65

### DIFF
--- a/glue/flagging_script_glue/flagging.py
+++ b/glue/flagging_script_glue/flagging.py
@@ -696,7 +696,8 @@ def z_normalize(df: pd.DataFrame, columns: list) -> pd.DataFrame:
                            as 'column_name_zscore'
     """
     for col in columns:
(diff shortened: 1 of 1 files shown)


dfsnow (approved):
Nice @wagnerlmichael, this looks good to me. Thanks for the quick fix.

PR #72, opened 2023-11-01
Switch standard deviation outlier detection boundaries to [2, 2]

wagnerlmichael:
Edit bounds based on decisions made from #65.


dfsnow (approved):
Great. See #65 for the context of this decision, but in sum: now that prices are logged, we'll be detecting far fewer sales on the high end using `[2, 3]` boundaries. Using `[2, 2]` puts us at _roughly_ the same percentage and number of outliers as using an unlogged `[2, 3]` set of bounds.

PR #73, opened 2023-11-02
Change params table to utilize athena arrays

### LINKED ISSUE #70: Edit col types to utilize athena array types
Currently the following columns in the `sale.parameter` table are list python objects written to strings:
- `iso_forest_cols`
- `res_stat_groups`
- `condo_stat_groups`
- `dev_bounds`
- `ptax_sd`

We need to utilize the existing athena array types to store these data instead of changing them to strings. This fits the data much better.

wagnerlmichael:
Edited `modify_dtypes` in order to fix the athena dtype overwrite. The `sale.parameter` table now contains the correct dtypes. 

Closes #70

### DIFF
--- a/glue/sales_val_flagging.py
+++ b/glue/sales_val_flagging.py
@@ -499,10 +499,6 @@ def modify_dtypes(df):
         df: df of standardized dtypes
     """
 
-    # Convert object columns to string
-    for col in df.select_dtypes("object").columns:
-        df[col] = df[col].astype("string")
-
     # Convert Int64 columns to int64
     for col in df.select_dtypes("Int64").columns:
         df[col] = df[col].astype("int64")


glue/sales_val_flagging.py
```diff
@@ -499,10 +499,6 @@ def modify_dtypes(df):
         df: df of standardized dtypes
     """
 
-    # Convert object columns to string
-    for col in df.select_dtypes("object").columns:
-        df[col] = df[col].astype("string")
-
```
wagnerlmichael, line 505: This is what was overwriting the array type conversation. I changed it and tested the an initial and manual run (writing both of the parameter dfs to athena), and after running the crawler it seems like there were no errors.

 I believe that the `Int64` conversion below is where the major error was (what this function was written for). I will double check this during a test final run but I'm fairly sure this will work.

dfsnow: Sounds good! Hopefully this is enough.

dfsnow (approved):
Looks good!

PR #79, opened 2023-12-04
Add standard deviation group data in group mean table

### LINKED ISSUE #78: Add standard deviation group data in group mean table
We currently do not collect standard deviation for the groups in our group mean table. We need to add this in order to be able to back track exactly why a sale was flagged in terms of price deviation.

wagnerlmichael:
We add standard deviation in price and price per sqft to the `group_mean` table so that we can back out the how a sale was flagged in terms of price.

If this looks good - I can wipe all of the tables, re-run the initial flagging script, and update the athena schema, if that seems like the best move.

Closes #78


dfsnow (commented):
@wagnerlmichael This looks good! But let's see the output in the `ci_` database before merging.

wagnerlmichael:
group_mean_1
group_mean_2

dfsnow (approved):

PR #87, opened 2024-01-18
Fix condo ptax flags, fix recent month ingest in glue script, add ptax directional flag

### LINKED ISSUE #86: Ptax flags not functioning with condos
Currently ptax flags are not attached to condo sales.

wagnerlmichael:
This PR does three things

- Adds directional flag to `sv_outlier_type` for ptax indicator
- Fixes a problem where the sales in the most recent month were getting clipped to the first day of that month in the ingest
- Makes sure condo sales get ptax flags

These changes have been tested on dev glue job resources.

Closes #86

### DIFF
--- a/glue/sales_val_flagging.py
+++ b/glue/sales_val_flagging.py
@@ -84,7 +84,7 @@ def add_rolling_window(df, num_months):
     return df
 
 
-def ptax_adjustment(df, groups, ptax_sd):
+def ptax_adjustment(df, groups, ptax_sd, condos: bool):
     """
     This function manually applies a ptax adjustment, keeping only
     ptax flags that are outside of a certain standard deviation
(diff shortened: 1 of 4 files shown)


glue/sales_val_flagging.py
```diff
@@ -102,12 +102,35 @@ def ptax_adjustment(df, groups, ptax_sd):
...
+        )
+
+        # Determine the ptax direction
+        conditions = [
+            (df[f"sv_price_deviation_{group_string}"] >= ptax_sd[1]),
+            (df[f"sv_price_deviation_{group_string}"] <= -ptax_sd[0]),
+        ]
+
+    directions = ["High", "Low"]
+    df["ptax_direction"] = np.select(conditions, directions, default=np.nan)
```
wagnerlmichael, line 133: I added this when working with @ccao-jardine. It splits PTAX by high or low outlier in `sv_outlier_type`. Nothing functionally changes, but one more value type will be added in `sv_outlier_type`. I added it to this PR because I think it makes sense to include as we are probably going to have to do a full re-run anyway to re-flag these condos, even if we don't go with a overall methodology change.

glue/sales_val_flagging.py
```diff
@@ -102,12 +102,35 @@ def ptax_adjustment(df, groups, ptax_sd):
 
     group_string = "_".join(groups)
 
-    df["ptax_flag_w_deviation"] = df["ptax_flag_original"] & (
-        (df[f"sv_price_deviation_{group_string}"] >= ptax_sd[1])
-        | (df[f"sv_price_deviation_{group_string}"] <= -ptax_sd[0])
-        | (df[f"sv_price_per_sqft_deviation_{group_string}"] >= ptax_sd[1])
-        | (df[f"sv_price_per_sqft_deviation_{group_string}"] <= -ptax_sd[0])
-    )
+    if condos == False:
```
wagnerlmichael, line 105: The reason the ptax flags weren't being assigned in condos is that there wasn't a conditional block that handles the different groups. This means that condo data used the res groups, resulting in na values for the conditional statements that the ptax outliers are built upon.

glue/sales_val_flagging.py
```diff
@@ -701,7 +727,7 @@ def write_to_table(df, table_name, s3_warehouse_bucket_path, run_id):
     NA_Dates AS (
         SELECT
             MIN(DATE_TRUNC('MONTH', sale.sale_date)) - INTERVAL '{rolling_window_num_sql}' MONTH AS StartDate,
-            MAX(DATE_TRUNC('MONTH', sale.sale_date)) AS EndDate
+            MAX(date_add('day', -1, date_add('month', 1, DATE_TRUNC('MONTH', sale.sale_date)))) AS EndDate
```
wagnerlmichael, line 730: For our most recent sales data, it was clipping the ingest on the first of that month. This change makes sure we get all of the data from that month.

manual_flagging/initial_flagging.py
```diff
@@ -175,18 +175,29 @@
...
+    [df_res_flagged_updated_ptax, df_condo_flagged_updated_ptax]
+).reset_index(drop=True)
+
 # Finish flagging and subset to write to flag table
 df_to_write, run_id, timestamp = flg.finish_flags(
-    df=df_flagged_ptax,
+    df=df_flagged_ptax_merged,
     start_date=inputs["time_frame"]["start"],
     manual_update=False,
 )
```
wagnerlmichael, line 203: Separates out ptax creation into two different data frames so that we are able to use condo-specific groups.

glue/sales_val_flagging.py
jeancochrane: **[Nitpick, non-blocking]** Makes sense! A more Pythonic check here would be to use `condos is False`, or better yet, check the truthiness of the value:

```suggestion
    if not condos:
```

manual_flagging/manual_update.py
```diff
@@ -186,23 +186,33 @@
...
-df_flagged_merged = pd.concat(
-    [df_res_flagged_updated, df_condo_flagged_updated]
-).reset_index(drop=True)
+# Update the PTAX flag column with an additional std dev conditional w/ res groups
+df_res_flagged_updated_ptax = flg.ptax_adjustment(
+    df=df_res_flagged_updated,
+    groups=inputs["stat_groups"],
+    ptax_sd=inputs["ptax_sd"],
+    condos=False,
+)
```
jeancochrane, line 195: **[Thought, non-blocking]** It's a bummer that to change any of the flagging logic, you have to make nearly identical changes in three different scripts; not a great practice for long-term maintainability. Once this pipeline gets to a stable state, we should brainstorm ways that we can centralize the flagging logic.

jeancochrane (approved):
Looking great!

dfsnow (approved):
This looks good to me. @wagnerlmichael let's merge it and do another Glue run.

PR #90, opened 2024-01-23
Architecture change for variable methodology

wagnerlmichael:
This is a PR that is rather large in scope. It re-hauls the way we do flagging, these changes allow us to flag specific partitions of sales. In the prior version, we couldn't flag sales by tri and we couldn't assign different flagging methodologies to each tri or housing market type. With these changes, we are able to do both of these things. We can choose which tris, which housing types, and which methods to flag for. This set of filters and methods can be selected in the `manual_flagging/yaml/inputs.yaml` config file.

We have also combined the `manual_flagging/initial_flagging.py` and the `manual_flagging/manual_update.py` into one file - `manual_update/flagging.py`. The recurring glue job [still needs to be refactored.](https://github.com/ccao-data/model-sales-val/issues/92).

We also have a `migrations/` directory where `transform_migrate.py` lives. When we did two recent sales val runs (https://github.com/ccao-data/model-sales-val/commit/f746297d0ae49760754e56902ea241d3f327f253 and https://github.com/ccao-data/model-sales-val/commit/6d9c1277c1006ab6174ffb1897f56b6adb4c9ee9), we had to save metadata to a separate location for the time being. The `transform_migrate` script performs a data migration. In this data migration we are changing the structure of the metadata and not the flags themselves. The data migration does multiple things.

- Moves the locally saved metadata tables (from https://github.com/ccao-data/model-sales-val/commit/f746297d0ae49760754e56902ea241d3f327f253 and https://github.com/ccao-data/model-sales-val/commit/6d9c1277c1006ab6174ffb1897f56b6adb4c9ee9) to prod.
- Makes a number of column adjustments and changes to the tables. 
    - The `group` columns in both `sale.flag` and `sale.group_mean` now have an additional `market` classifier added to them. This was necessary in the case of the same groups across two different housing market types.
    - Edited multiple columns in `sale.parameter`, where many now take string representations of json data directly from the config file.


manual_flagging/yaml/inputs_initial.yaml
```diff
@@ -2,11 +2,92 @@
...
+        - "rolling_window"
+        - "geography_split"
+      class_filter: ["297", "299", "399"]
+  og_mansueto:
+    res_single_family:
+      columns:
+        - "rolling_window"
+        - "township_code"
+        - "class"
+      class_filter: [
```
wagnerlmichael, line 54: I think I need to bring these `class_filter` lists out and incorporate them with `housing_run_type`

manual_flagging/initial_flagging.py
```diff
@@ -121,14 +130,167 @@
...
+        if "condos" in df_name:
+            pass
+
+    # Add the edited or unedited dataframe to the new dictionary
+    dfs_to_rolling_window[df_name] = df_copy
+
+
+# Make rolling window
+#
+#
```
wagnerlmichael, line 293: So far I have just implemented filters and feature creation. My next step would be to move to rolling_window creation and the actual flagging.

manual_flagging/initial_flagging.py
```diff
@@ -121,14 +130,167 @@
...
+# Make correct filters
+#
+
+# Filter to correct tris
+tri_stat_groups = {
+    tri: method
+    for tri, method in inputs["tri_stat_groups"].items()
+    if tri in inputs["run_tri"]
+}
```
wagnerlmichael, line 147: `inputs` is the yaml object

glue/flagging_script_glue/flagging.py
```diff
@@ -130,12 +124,18 @@ def iso_forest(
...
         [(df["sv_anomaly"] == -1), (df["sv_anomaly"] == 1)],
         ["Outlier", "Not Outlier"],
         default="Not Outlier",
     )
 
+    # Restore original values for encoded columns
+    for group, le in label_encoders.items():
+        df[group] = le.inverse_transform(df[group])
+
+    # Reset index
```
wagnerlmichael, line 138: This change to `iso_forest` is already merged into master, details [here](https://github.com/ccao-data/model-sales-val/pull/93#discussion_r1470029355).

glue/flagging_script_glue/flagging.py
```diff
@@ -849,15 +849,15 @@ def outlier_type(df: pd.DataFrame, condos: bool) -> pd.DataFrame:
...
+            # "Anomaly (high)",
             "High price swing",
             "High price (raw & sqft)",
             "High price (raw)",
             "High price (sqft)",
             "Home flip sale (low)",
             "Family sale (low)",
             "Non-person sale (low)",
-            "Anomaly (low)",
+            # "Anomaly (low)",
```
wagnerlmichael, line 860: The commented out iso_forest code is just to enable faster iteration since it is the longest part of the flagging proccess.

wagnerlmichael:
## Summary

The purpose of this review is two-fold.
1. Get feedback early enough so that if it makes sense to re-do everything I'm not too far deep into the dev process such that it is a waste of time.
2. Get advice on a number of architecture decision.

## Motivation

## Main Logic

## To do

## Outstanding Questions

glue/sales_val_flagging.py
```diff
@@ -427,19 +427,8 @@ def finish_flags(df, start_date, manual_update):
...
-        "rolling_window": lambda df: pd.to_datetime(
-            df["rolling_window"], format="%Y%m"
-        ).dt.date,
-    }
-
-    if not manual_update:
-        dynamic_assignment["version"] = 1
-
     # Finalize to write to sale.flag table
-    df = df[cols_to_write].assign(**dynamic_assignment).reset_index(drop=True)
```
wagnerlmichael, line 442: This code was actually redundant. It was assigning versions of 1 or the absence of a version column depending on the `manual_update` boolean. Either way, the existing data got incremented by 1 anyway if we were doing a manual update.

manual_flagging/initial_flagging.py
```diff
@@ -121,153 +140,373 @@
...
 
-# Separate res and condo sales based on the indicator column
-df_res = df[df["indicator"] == "res"].reset_index(drop=True)
-df_condo = df[df["indicator"] == "condo"].reset_index(drop=True)
+# Filter to correct tris
+tri_stat_groups = {
+    tri: method
+    for tri, method in inputs["tri_stat_groups"].items()
+    if tri in inputs["run_tri"]
+}
```
wagnerlmichael, line 148: This narrows down the scope of what we plan to flag. If we only have tri 2 and 3 in the yaml list, this will use that to connect these two tris to the run methodology (`current` or `og_mansueto`) in the `tri_stat_groups` object, which acts as a mapping to tri and run methodology.

manual_flagging/initial_flagging.py
```diff
@@ -121,153 +140,373 @@
...
+    df["nbhd"] = df["nbhd"].astype(int)
+    df = pd.merge(df, df_new_groups, on="nbhd", how="left")
+
+# - - - - - - -
+# Make correct filters and set up dictionary structure
+# - - - - - - -
+
+dfs_to_feature_creation = {}  # Dictionary to store DataFrames
+
+for tri, method in tri_stat_groups.items():
```
wagnerlmichael, line 170: This for loop looks at the `inputs_intitial.yaml` and creates a dictionary structure that is used and added on in the rest of the script.

manual_flagging/initial_flagging.py
```diff
@@ -121,153 +140,373 @@
...
-    s3_warehouse_bucket_path=os.getenv("AWS_S3_WAREHOUSE_BUCKET"),
-    run_id=run_id,
-)
+    dfs_to_write[df_name]["df_metadata"] = df_metadata
+
+# - - - -
+# Write tables
+# - - - -
+
+for df_name, df in dfs_to_write.items():
```
wagnerlmichael, line 479: In the old structure, the order was as follows:
- create table
- write table
- create table
- write table

I think this new change is better:
- create all tables
- write all tables

When we had the out of memory error, it happened in-between writing different tables, which resulted in some meta data tables being written to prod, and some being lost forever, which meant that we had to manually delete the table and re-run. With this change, we create all of the table first, so if there happens to be an error in creating any of them, non of them will write, so we wouldn't have to manually nuke prod tables.

manual_flagging/yaml/inputs_initial.yaml
```diff
@@ -1,28 +1,114 @@
...
+  1: "current"
+  # north
+  2: "og_mansueto"
+  # south
+  3: "og_mansueto"
+
+stat_groups:
+  current:
+    res_single_family:
+      columns:
```
wagnerlmichael, line 35: Stat groups is first split into `current` and `og_mansueto` keys, which further split into more supplementary info for flagging in each respective methodology.

manual_flagging/yaml/inputs_initial.yaml
```diff
@@ -1,28 +1,114 @@
...
+    "sv_sale_dup_counts"
+  ]
+  condos: [
+    "meta_sale_price",
+    "sv_days_since_last_transaction",
+    "sv_cgdr",
+    "sv_sale_dup_counts"
+  ]
+
+housing_run_type_filters:
```
jeancochrane, line 83: **[Nitpick, non-blocking]** Perhaps we could make it clear from the attribute name that this is a group of classes -- maybe something like `housing_market_class_codes`?

manual_flagging/yaml/inputs_initial.yaml
```diff
@@ -1,28 +1,114 @@
...
+        - "township_code"
+
+iso_forest:
+  res: [
+    "meta_sale_price",
+    "sv_price_per_sqft",
+    "sv_days_since_last_transaction",
+    "sv_cgdr",
+    "sv_sale_dup_counts"
+  ]
```
jeancochrane, line 75: **[Nitpick, non-blocking]** A more YAMLific (not a real term) way of defining an array of values:

```suggestion
  res:
    - "meta_sale_price"
    - "sv_price_per_sqft"
    - "sv_days_since_last_transaction"
    - "sv_cgdr"
    - "sv_sale_dup_counts"
```

manual_flagging/yaml/inputs_initial.yaml
jeancochrane: Also: Are these values saved to the correct metadata/parameter table yet? If not, maybe we should think about doing that so that it matches the way the other configs are treated.

jeancochrane (commented):
This is looking great so far! I think you're definitely on the right track, and it's starting to get close to ready for a final review. I only reviewed things from a high level, and will leave more fine-grained review until the design is totaly settled.

To respond to your questions:

> 1. Currently each row of the parameter (or metadata) table represents a specific tri/market/methodology combination. However, it isn't explicitly stated in metadata tables . I can't quite remember all of the options discussed with @jeancochrane. I was thinking I could just programmatically output a string - something like "city-tri_res_single_family_current", which would be a new column in sale.parameter called flag_model_specification or something like that. I was also thinking of outputting something from the inputs_initial.yaml in a json format, but I couldn't settle on a good way to do it.

I left a comment on the code that should hopefully address this! In short, my suggestion is we try outputting parameter/metadata values as JSON strings so that we can just dump the entire config once for every run. Happy to talk this through on a call if it would help.

> 2. The bin creation from inputs_yaml also are not saved in any of the output tables, but they are preserved by way of commit hash. Is this sufficient? Does the generalizability/specifics trade-off make sense here?


I think another advantage to dumping parameter/metadata values as JSON strings is that we would include bins by default, since I believe they're defined in `stat_groups`. But if for some reason we can't go in that direction, let's chat about alternative options.

> 3. I currently [read in a messily named excel file](https://github.com/ccao-data/model-sales-val/blob/3b79930c0703d6e09a42808b66cc39d6216c6a14/manual_flagging/initial_flagging.py#L150-L162) to create the geography groups for the current methodology. Obviously we should change this. I was thinking I could build in a relative path call, and build the path to the file into the inputs_initial.yaml file such that we could have multiple grouping data for each method. I could also make a designated directory for this type of data. How does this sound?

I like both of those ideas! I wouldn't worry too much about getting this detail perfect as part of this PR, since it'll be easy to refactor in the future and doesn't affect the metadata/parameter tables.

> 4. Should I put everything into callable functions and use click like mentioned https://github.com/ccao-data/model-sales-val/issues/92. I was thinking of merging manual update and initial into script in this branch, and then leaving the recurring/glue job for https://github.com/ccao-data/model-sales-val/issues/92.

I think it makes sense to focus this PR on consolidating initial and manual flagging around one script and one YAML file, and then waiting for a follow-up PR to refactor such that it can be run on Glue (or whatever compute platform we decide to use going forward).

manual_flagging/yaml/inputs_initial.yaml
wagnerlmichael: Going to start using "YAMLific" in everyday life

manual_flagging/initial_flagging.py
```diff
@@ -121,153 +140,375 @@
...
+    tri: method
+    for tri, method in inputs["tri_stat_groups"].items()
+    if tri in inputs["run_tri"]
+}
 
-# Create condo stat groups. Condos are all collapsed into a single class,
-# since there are very few 297s or 399s
-condo_stat_groups = inputs["stat_groups"].copy()
-condo_stat_groups.remove("class")
+# Handle current methodology data manipulation if needed (temporary, need to find a better way)
```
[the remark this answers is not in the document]
wagnerlmichael: Given that the bin partitions are different (at least currently in the `current` methodology) depending on which property type, there would have to be a conditional check somewhere no matter what since the bins would have to be specific to the filtered data. I could theoretically make two different column for `res_multi_family` and `res_single_family`, but I would still have to do a conditional check to properly allocate the columns.

You'll have to let me know if you can think of a better way to do this. One potential positive of this feature creation loop is that it is its own section where we can see for which data we are doing which extra splits, potentially making it slightly more readable? Not sure.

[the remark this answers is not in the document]
wagnerlmichael: I think that makes sense, I'll change it.

manual_flagging/initial_flagging.py
```diff
@@ -121,153 +140,375 @@
...
+                    "res" if "res" in market else "condos"
+                ],
+                "condos_boolean": market == "condos",
+            }
+
+        elif method == "og_mansueto":
+            # Collect all mansueto tris
+            mansueto_tris_to_flag = [
+                str(key)
+                for key, value in tri_stat_groups.items()
```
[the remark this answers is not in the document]
wagnerlmichael: I might be misunderstanding your point but I'll try to further explain the idea behind the code. The reason that we are collecting mansueto tris here:
```python
mansueto_tris_to_flag = [
    str(key)
    for key, value in tri_stat_groups.items()
    if value == "og_mansueto"
]
```
is that with the og_mansueto method, the groupings (rolling_window, class, township for res // rolling_window, township for condos) work no matter how many tris are included. This is is due to the fact that `og_mansueto` wasn't built to flexibly flag sales based on housing type, and that townships discretely make up a tri. So if we have tri 2 and tri 3 configured to `og_mansueto`, I thought to combine them into one data frame instead of iterating over each tri separately:

```python
 triad_code_filter = ~df["triad_code"].isin(mansueto_tris_to_flag)

df_res_og_mansueto = df[(df["indicator"] == "res") & triad_code_filter]
df_condo_og_mansueto = df[(df["indicator"] == "condo") & triad_code_filter]
```

manual_flagging/initial_flagging.py
```diff
@@ -121,153 +140,375 @@
...
+        iso_forest_cols=dfs_to_write[df_name]["iso_forest_cols"],
+        stat_groups=dfs_to_write[df_name]["columns"],
+        dev_bounds=inputs["dev_bounds"],
+        ptax_sd=inputs["ptax_sd"],
+        rolling_window=inputs["rolling_window_months"],
+        date_floor=inputs["time_frame"]["start"],
+        short_term_thresh=flg_model.SHORT_TERM_OWNER_THRESHOLD,
+        min_group_thresh=inputs["min_groups_threshold"],
+        run_id=dfs_to_write[df_name]["run_id"],
+    )
```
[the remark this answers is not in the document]
wagnerlmichael: I think the group means could very simply just be concatenated together 

I like that idea, but I *think* there are downsides to both. One large motivation for these metadata tables is the ability to have a granular look at how a sale came to be flagged, whether it be historically or as the most recent iteration. In the current model that I'm thinking about (one `run_id` per flagged group of data, one row per parameter/metadata table), I think it is slightly easier to do this. 

As a theoretical example: Lets say there is a sale I need to look at, and I need to figure out why it was flagged as an outlier. First I look at the `run_id`, then I reference `run_id` the `sale.parameter` table to check the grouping columns for explainability 

**comment not finished**

manual_flagging/yaml/inputs_initial.yaml
wagnerlmichael: > **[Nitpick, non-blocking]** Perhaps we could make it clear from the attribute name that this is a group of classes -- maybe something like `housing_market_class_codes`?

Done!

> so that it matches the way the other configs are treated.

Could you elaborate a little bit here?

manual_flagging/initial_flagging.py
```diff
@@ -121,153 +140,375 @@
...
+
+for df_name, df in dfs_to_feature_creation.items():
+    # Make a copy of the data frame to edit
+    df_copy = df["df"].copy()
+
+    # print(f"{df_name}")
+    if "current" in df_name:
+        print("current")
+
+        if "res_single_family" in df_name:
```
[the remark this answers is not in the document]
wagnerlmichael: Going to work out a solution for this and link a commit for you to check out

manual_flagging/yaml/inputs_initial.yaml
jeancochrane: > Could you elaborate a little bit here?

Yeah, sorry for the lack of clarity! I just meant to say that if we're not yet persisting these groupings to the metadata or parameter table as we do with e.g. `stat_groups`, we should probably think about doing so.

manual_flagging/initial_flagging.py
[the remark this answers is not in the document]
jeancochrane: I like the way you're thinking about the tradeoffs here! But I'm not totally convinced that the multi-row solution incurs substantially greater precision, and I'm not yet sure how important that precision actually is.

My understanding is that in the multi-row scenario, we would still need to figure out the tri and market type for a sale before we knew what stat groupings applied to it, since in neither solution is there a direct reference between a sale and the parameters that applied to it (we only have `run_id`, which would not uniquely identify a `parameter` row under the multi-row solution, and would not uniquely identify an entry in the nested `stat_groups` data structure under the single-row solution). But maybe I'm misunderstanding your proposal for a unique key; happy to discuss this more.

My other reason for leaning towards a single-row solution is that I don't yet have a ton of confidence that it's mission-critical for these tables to be as precise as possible. As long as we can quickly diagnose the parameters that generated an individual sale, we should be fine; both of these solutions fit that bill in my book, in which case I lean toward the simpler solution since I expect it to be easier to maintain going forward (less data and a simpler data model).

Perhaps we could discuss on a call today or tomorrow?

manual_flagging/initial_flagging.py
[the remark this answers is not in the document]
jeancochrane: Gotcha! I think it's reasonable to want to handle the legacy method in a way that more closely reflects the structure of its data. In that case, I would advocate for splitting the `tri_stat_groups` loop by method, something like:

```python
# Split tris into groups according to their flagging methods
current_tris = [tri for tri, method in tri_stat_groups.items() if method == "current"]
og_mansueto_tris = [tri for tri, method in tri_stat_groups.items() if method == "og_mansueto"]

# Flag tris that use the most contemporary flagging method
for tri in current_tris:
    for market in inputs["housing_run_type"]:
        ...

# Flag tris that use the legacy method developed by the Mansueto Institute
triad_code_filter = ~df["triad_code"].isin(og_mansueto_tris)
...
```

That way we can avoid doing the confusing thing of iterating the `tri_stat_groups` object inside a loop that is already operating on its elements.

manual_flagging/initial_flagging.py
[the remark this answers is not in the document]
wagnerlmichael: https://github.com/ccao-data/model-sales-val/commit/f06786c043448a1e1a48c9bc5ecbaff8f2e6321b

manual_flagging/initial_flagging.py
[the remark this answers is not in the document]
wagnerlmichael: Implemented this! I scrapped the loop for `og_mansueto` method.

wagnerlmichael:
## Docs for another review - 2/15/2024 (in progress)

manual_flagging/yaml/inputs_initial.yaml
wagnerlmichael: Done.

glue/sales_val_flagging.py
```diff
@@ -523,15 +512,47 @@ def modify_dtypes(df):
...
+        "run_filter",
+        "iso_forest_cols",
+        "stat_groups",
+        "tri_stat_groups",
+        "housing_market_class_codes",
+    ]
+    for col in string_columns:
+        if col in df.columns:
+            df[col] = df[col].astype(str)
+
```
wagnerlmichael, line 544: This change helps to standardize dtypes, it is built for the `sale.parameter` table. This addition solves two problems:

- String encoded json columns. This is a hard check that all of the columns in `string_columns` are changed to string type before being written to athena. This check helps to make sure athena dtypes are consistent so there are no issues when building the tables
- Nested `array<double>` problems. There was a problem with the `ptax_sd` column data. In some runs the values in the python list that gets written to athena are float, and in others the values are int. When a python list is written to athena, the resulting dtype is `array<$NUMERIC_DTYPE>`.  Since athena can't handle compound types, it tries to translate the entire `array` to either double or int. This resulted in different parquets having a different athena dtype which cause hive errors upon using athena to query the table. The code in the function manually changes all of the elements within the lists to float, so that no matter what athena will interpret them as `array<double>`
- Another problem was found in the process of reading prod data, manipulating it, and re-writing it. The fix for the prior issue works as we look inside a list and change the dtype of the elements inside the list. A problem with this approach was that when reading existing parquet data from prod, these `array` columns are read back into python as numpy arrays. Manually changing the elements of a list does not work for the numpy array data structure, so we manually change any numpy array to a list so that the int --> float conversion works in all scenarios.

transform_migrate.py
```diff
@@ -0,0 +1,366 @@
...
+    A dictionary of DataFrames keyed by their names.
+    """
+    # List all parquet files in the specified S3 path
+    parquet_files = wr.s3.list_objects(
+        os.path.join(os.getenv("AWS_S3_WAREHOUSE_BUCKET"), "sale", table),
+        suffix=".parquet",
+    )
+
+    # Initialize a dictionary to hold the dataframes
+    dfs = {}
```
wagnerlmichael, line 43: This script handles each of the transformations by `table` first (`flag/parameter/group_mean/metadata`). We transform the existing prod data, and we transform the locally saved metadata from our recent res and condos update.

transform_migrate.py
```diff
@@ -0,0 +1,366 @@
...
+
+    # Loop through the parquet files, read each into a DataFrame, and store it in the dictionary
+    for file in parquet_files:
+        # Extract a meaningful name part from the file path for use in the DataFrame variable name
+        name_part = file.split("/")[-1].split(".")[
+            0
+        ]  # Adjust this as necessary based on your file naming convention
```
wagnerlmichael, line 50: Gives the names of the data frames in python the same name as the file in S3 (`run_id`)

transform_migrate.py
```diff
@@ -0,0 +1,366 @@
...
+dfs_sale_flag = read_parquets_to_dfs("", "flag")
+
+# OG initial run
+# 'res_og_mansueto', 'condos_og_mansueto'
+dfs_sale_flag["2024-01-19_18:46-clever-boni"]["group"] = dfs_sale_flag[
+    "2024-01-19_18:46-clever-boni"
+]["group"].apply(
+    lambda value: value + "-res_og_mansueto"
+    if value.count("_") == 2
+    else (value + "-condos_og_mansueto" if value.count("_") == 1 else value)
```
wagnerlmichael, line 76: This retroactively appends the `housing_market_type` to the `group` columns. This works for the following reason.

In `og_mansueto` for res, we had 3 groups (`township_code`, `rolling_window`, and `class`), appended by `_`. In the condos groups we had 2 groups (`rolling_window` and `township_code`), appended by `_`. Thus we know the number of `_` per housing market type and we assign the suffix based on that.

transform_migrate.py
```diff
@@ -0,0 +1,366 @@
...
+)
+
+# Update res run
+dfs_sale_flag["2024-01-29_14:40-pensive-rina"]["group"] = dfs_sale_flag[
+    "2024-01-29_14:40-pensive-rina"
+]["group"].apply(
+    lambda value: value + "-res_single_family"
+    if "40_years" in value
+    else (value + "-res_multi_family" if "20_years" in value else value)
```
wagnerlmichael, line 85: This works because in the res update we only had two housing market groups flagged, `res_single_family` and `res_multi_family`, and we know that for the `res_multi_family` sales we use 20 years for binning, whereas for the `res_single_family`, we used 40 years.

transform_migrate.py
```diff
@@ -0,0 +1,366 @@
...
+)
+
+# Update new condos run
+dfs_sale_flag["2024-02-01_12:24-nifty-tayun"]["group"] = (
+    dfs_sale_flag["2024-02-01_12:24-nifty-tayun"]["group"] + "-condos"
```
wagnerlmichael, line 90: We can update condos to this entire run because the condos update only contained the `condos` housing market type to update.

transform_migrate.py
```diff
@@ -0,0 +1,366 @@
...
+)
+
+write_dfs_to_s3(dfs_sale_flag, os.getenv("AWS_TEST_ARCH_CHANGE_BUCKET"), "flag")
+
+# - - - - - -
+# Adjust parameter tables
+# - - - - - -
```
wagnerlmichael, line 97: In these parameter tables, the string type allowed me to have some flexibility in imputing prior run data into the structure that all of the new runs will have. For example, currently in the `stat_groups` columns, there is a long list of json string data written which contains bin information, current vs res, etc. For this og run, I impute this string:

``` python
{
        "og_mansueto": {
            "res": {"columns": ["rolling_window", "township_code", "class"]},
            "condos": {"columns": ["township_code", "rolling_window"]},
        }
    }
```

The json structure is different that the new existing structure for this column, so I decided to just simplify it to contain the relevant information. The drawback here is that the full json structure isn't consistent accross different run_id's. This can be observed in the `ci_model_sales_val_89_architecture_change_for_variable_methodology_sale.parameter` table. One alternative to the current method of "only putting what is necessary" would be to have the existing structure for every run_id and leaving lots of key values blank. You'll have to let me know what you think.

manual_flagging/flagging.py
```diff
@@ -0,0 +1,517 @@
...
+    assert "run_tri" in inputs, "Missing key: 'run_tri'"
+    assert set(inputs["run_tri"]).issubset(
+        {1, 2, 3}
+    ), "run_tri can only contain 1, 2, 3"
+    assert len(inputs["run_tri"]) == len(
+        set(inputs["run_tri"])
+    ), "Duplicate values in 'run_tri'"
+
+except AssertionError as e:
+    print(f"Error: {e}")
```
wagnerlmichael, line 44: This chunk is used to validate proper values are specified in the run filters in the `housing_market_type` and `run_tri` lists of the `inputs.yaml` file. It checks for valid values and duplicates.

jeancochrane (commented):
Looking great! I have a couple open questions below, so perhaps let's sync up on a call before we merge to make sure everything is buttoned up?

To respond to your questions above:

> I'm not sure the best way to handle this migration script in terms of its use in production. Do we run it as a one off and then delete it?

My recommendation would be to store it in a new subdirectory called `migrations/` and name it something like `2024_02_21_update_initial_group_metadata.py`. That way we'll have a log of the migration we ran but we can keep it in a place where it's clearly marked as legacy code.

> Perhaps we can approve the data in the test tables, nuke the prod data (or move it elsewhere so that we can use it in case we missed something), and then just copy the test parquets and place them in the prod buckets.

I like this plan! And I definitely think we should move the old prod data somewhere else so that we can revert back to it in case anything goes wrong. Maybe it would make sense to pair on this migration once it's ready to go?

transform_migrate.py
wagnerlmichael: Oooh I like this idea. I'll see if I can draft a prototype.

glue/flagging_script_glue/flagging.py
```diff
@@ -47,7 +48,7 @@ def go(
     print("create_stats() done")
     df = string_processing(df)
     print("string_processing() done")
-    df = iso_forest(df, groups, iso_forest_cols)
+    # df = iso_forest(df, groups, iso_forest_cols)
```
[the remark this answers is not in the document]
wagnerlmichael: Correct and correct

glue/sales_val_flagging.py
```diff
@@ -563,8 +584,9 @@ def get_parameter_df(
     earliest_sale_ingest = df_ingest.meta_sale_date.min()
     latest_sale_ingest = df_ingest.meta_sale_date.max()
     iso_forest_cols = iso_forest_cols
-    res_stat_groups = res_stat_groups
-    condo_stat_groups = condo_stat_groups
+    stat_groups = stat_groups
+    tri_stat_groups = tri_stat_groups
+    housing_market_class_codes = housing_market_class_codes
```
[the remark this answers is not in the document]
wagnerlmichael: Yeah, this sounds good. I remember for some reason it helped me avoid an error, but it was probably user error since `x = x` doesn't do anything.

group_data/res_condos_nbhd_groups_2024.xlsx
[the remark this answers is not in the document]
wagnerlmichael: Definitely a better name!

manual_flagging/flagging.py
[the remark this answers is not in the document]
wagnerlmichael: Correct, I consolidated manual flagging and initial flagging into this singular script.

manual_flagging/flagging.py
```diff
@@ -0,0 +1,517 @@
...
+        "iso_forest_cols": inputs["iso_forest"]["condos"],
+        "condos_boolean": True,
+        "market": "condos_og_mansueto",
+    }
+
+# - - - - - -
+# Make rolling window
+# - - - - - -
+
+dfs_to_flag = copy.deepcopy(dfs_to_rolling_window)
```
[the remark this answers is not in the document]
wagnerlmichael: Going to take a stab at replying here before we hop on a call. Let me know if I misunderstood your question. I see two drawbacks with instantiating an empty dictionary instead of using `deepcopy` here. 

The first is that copying the dict gives us all of the auxiliary data for each dataframe.
```python
# - - - - -
# Flag Sales
# - - - - -

dfs_flagged = copy.deepcopy(dfs_to_flag)

for df_name, df in dfs_to_flag.items():
    print(f"\nFlagging sales for {df_name}")
    df_copy = df["df"].copy()
    df_copy = flg_model.go(
        df=df_copy,
        groups=tuple(df["columns"]),
        iso_forest_cols=df["iso_forest_cols"],
        dev_bounds=tuple(inputs["dev_bounds"]),
        condos=df["condos_boolean"],
    )

    # Add the edited or unedited dataframe to the new dictionary
    dfs_flagged[df_name]["df"] = df_copy
```

manual_flagging/flagging.py
```diff
@@ -0,0 +1,517 @@
...
+                    labels=bldg_age_labels,
+                )
+                dfs_to_rolling_window[key]["df"] = df_copy
+
+        if tri == 2:
+            # Currently have not developed a new method for tri 2
+            pass
+        if tri == 3:
+            # Currently have not developed a new method for tri 3
+            pass
```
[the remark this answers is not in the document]
wagnerlmichael: Implemented a version of that here: https://github.com/ccao-data/model-sales-val/pull/90/commits/0a4d7db9231c3f5633f98fe11162f448541a8b82

manual_flagging/flagging.py
```diff
@@ -0,0 +1,517 @@
...
+
+for df_name, df in dfs_to_finalize.items():
+    df_group_mean = flg.get_group_mean_df(
+        df=dfs_flagged[df_name]["df"],
+        stat_groups=df["columns"],
+        run_id=run_id,
+        condos=df["condos_boolean"],
+    )
+    market_value = df["market"]
+    df_group_mean["group"] = df_group_mean["group"].astype(str) + "-" + market_value
```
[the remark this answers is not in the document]
wagnerlmichael: Tried this for about 2 hours and got stumped. It requires a decent amount of changes in the flagging pipeline so I'm thinking about skipping it for now. It might be worth making an issue about this though.

manual_flagging/flagging.py
```diff
@@ -0,0 +1,517 @@
...
+    """
+    Modify the 'group' column by appending '-market_value', this is done
+    to make sure that a two different groups with the same run_id won't
+    be returned with the same value. For example, if res and condos have the 
+    same column groupings, joining the group column from sale.flag to sale.group_mean
+    by 'group' and 'run_id' could potentially return two groups. This market type
+    append fixes that. This is also added in the group_mean data.
+    """
+    market_value = df["market"]
+    df_copy["group"] = df_copy["group"].astype(str) + "-" + market_value
```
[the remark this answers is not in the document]
wagnerlmichael: Related to this comment https://github.com/ccao-data/model-sales-val/pull/90#discussion_r1502752892

manual_flagging/flagging.py
```diff
@@ -21,17 +21,25 @@
...
 
+# Check housing_market_type for mutual exclusivity
+assert not (
+    (
+        "res_single_family" in inputs["housing_market_type"]
+        or "res_multi_family" in inputs["housing_market_type"]
+    )
+    and "res_all" in inputs["housing_market_type"]
+), "Either 'res_single_family' and 'res_multi_family' can be included, or 'res_all', but not both"
+
```
wagnerlmichael, line 41: Check to make sure a subsection of `"res_all"` isn't selected along iwth `"res_all"`.

manual_flagging/yaml/inputs.yaml
```diff
@@ -12,9 +12,13 @@ manual_update: false
...
-# list supports 'res_single_family', 'res_multi_family', and/or 'condos'.
-# The run_tri van be 1, 2, and/or 3, and the values for the tri_stat_groups
-# can be "current" or "og_mansueto".
+# list supports 'res_single_family', 'res_multi_family', 'res_all,' and/or
+# 'condos'. If we have a methodology that flags all of res, we will use
+# 'res_all'. If we need to have a separate methodology for single and
+# multi-family, we can use  those respective values. This means that if res_all
+# is selected, we can't use either of the other res market values, as it would
+# be redundant. The run_tri van be 1, 2, and/or 3, and the values for the
+# tri_stat_groups can be "current" or "og_mansueto".
```
wagnerlmichael, line 21: This change is implemented in the potential case where we want to flag all of res with the same methodology.

manual_flagging/flagging.py
wagnerlmichael: Scrapped this change as I'm not sure it will be needed.

manual_flagging/yaml/inputs.yaml
wagnerlmichael: Scrapped this change as I'm not sure it will be needed.

glue/sales_val_flagging.py
```diff
@@ -427,6 +427,17 @@ def finish_flags(df, start_date):
...
+    dynamic_assignment = {
+        "run_id": run_id,
+        "rolling_window": lambda df: pd.to_datetime(
+            df["rolling_window"], format="%Y%m"
+        ).dt.date,
+    }
+
+    if not manual_update:
+        dynamic_assignment["version"] = 1
+
```
wagnerlmichael, line 440: I had removed this because I thought it was redundant but I was mistaken. This is here to manually assign version of 1 when `manual_update` is false.

manual_flagging/flagging.py
```diff
@@ -284,10 +284,10 @@ def create_bins_and_labels(input_list):
 
 if og_mansueto_tris:
     # Filter by triad code and market type
-    og_mansueto_filter = ~df["triad_code"].isin(og_mansueto_tris)
+    og_mansueto_filter = df["triad_code"].astype(int).isin(og_mansueto_tris)
 
-    df_res_og_mansueto = df[(df["indicator"] == "res") & triad_code_filter]
-    df_condo_og_mansueto = df[(df["indicator"] == "condo") & triad_code_filter]
+    df_res_og_mansueto = df[(df["indicator"] == "res") & og_mansueto_filter]
+    df_condo_og_mansueto = df[(df["indicator"] == "condo") & og_mansueto_filter]
```
wagnerlmichael, line 290: Not sure how I missed this, but the filter operation was incorrect here, resulting in incorrect sales being flagged. This is fixed and tested now.

glue/flagging_script_glue/flagging.py
```diff
@@ -797,13 +800,13 @@ def outlier_type(df: pd.DataFrame, condos: bool) -> pd.DataFrame:
             "Home flip sale (high)",
             "Family sale (high)",
             "Non-person sale (high)",
-            "Anomaly (high)",
+            # "Anomaly (high)",
```
jeancochrane, line 803: **[Question, non-blocking]** Do we want `Anomaly (high)` to continue to be commented out?

migrations/transform_migrate.py
jeancochrane, line 1: **[Nitpick, non-blocking]** I think it'd be helpful to give this file a slightly more descriptive name, and to include a leading number in the filename (a la [Django migrations](https://docs.djangoproject.com/en/5.0/topics/migrations/#workflow)) so that we know the order the migrations were added on the off chance that we ever need to add more:

```
migrations/0001_update_metadata_structure_after_initial_run.py`
```

It's also probably worth adding a brief docstring to the top of the file explaining what it's for, something like:

```python
"""Migration to update the structure of the prod metadata following the initial flagging run.
This migration does not change any flags, but rather reads data from the prod bucket, transforms
it so that its metadata matches the most recent schema, and saves it to a new bucket where
it can be QCed before being copied back to the prod bucket. This migration was necessary in late
Feb 2024 since prod flags for the 2023 reassessment were computed before the metadata schema
was completely solidified."""
```

jeancochrane (approved):
A couple truly tiny thoughts (two file naming and comment suggestions) but you should feel free to merge once you've resolved them to your liking! Looking forward to updating the prod data tomorrow.

glue/flagging_script_glue/flagging.py
wagnerlmichael: Nope, thanks for catching.

Issue #89, opened 2024-01-22
Potential Architecture Change


wagnerlmichael:
If we decide to move forward with a new sales val strategy. We will need to overhaul config files and meta data output tables.

In terms of inputs we have a [yaml file](https://github.com/ccao-data/model-sales-val/blob/main/manual_flagging/yaml/inputs_initial.yaml) with columns for the groups we use
``` yaml
stat_groups: [
  "rolling_window",
  "township_code",
  "class"
]
```

We have a number of output tables that act as supplementary information to the `sale.flag` table. They are`sale.parameter`, `sale.group_mean`, and `sale.metadata`.

The `sale.parameter` data has the following columns:
- `run_id`
- `sales_flagged`
- `earliest_data_ingest`
- `latest_data_ingest`
- `iso_forest_cols`
- `res_stat_groups`
- `condo_stat_groups` (generated *in* the script by removing a column - potential problem here)
- `dev_bounds`
- `ptax_sd`
- `rolling_window`
- `date_floor`
- `short_term_owner_threshold`
- `min_group_thresh`

The `sale.group_mean` table is comprised of
- `group` - (a single string appended with a `_` separator, for example - `201703_27_295` (rolling_window_township_class) )
- `group_mean`
- `group_std`
- `group_sqft_std`
- `group_sqft_mean`
- `group_size`
- `run_id`

The `sale.metadata` table is comprised of 
- `run_id`
- `long_commit_sha`
- `short_commit_sha`
- `run_timestamp`
- `run_type`

### Groupings that aren't native columns

Right now, the yaml input as a column works nicely with the output structure. In the case of new potential splits. It won't be so clean.  

- New geographies currently ingest from an excel file 
    - Potential solutions
        - Build excel data into an existing table or view? 
- New age and size bins
    - Potential solutions
        - Binning done within sql query

### Update for only a tri
- This new methodology (for now) would only be run on 1 tri. We need to add that option
- Potential solution - add a yaml input as `tris` that could include 1-3

### Different groupings for single/multi-family + condos in the near future
- Single family groups would theoretically contain time window, geographies, size, and age, whereas multi-family would be the same but excluding a size grouping
- The same type of thing will happen when we go to run condo groupings. It is possible that we will decide on yet another distinct set of groups for condos. Currently, the condo groups are dynamically created by dropping an element of a list in the script, we should probably put this into an input config file.
- These changes will probably need to have new columns introduced in the output tables.

### Can old tables and new tables live together?
[...]

PR #95, opened 2024-01-31
Update condos flags based on new res grouping method

wagnerlmichael:
This PR updates the city tri condo flags with the new geography splits from res. This is an admittedly messy, temporary solution to update the flags, similar to the [res equivalent](https://github.com/ccao-data/model-sales-val/commit/f746297d0ae49760754e56902ea241d3f327f253). When the new [architecture updates](https://github.com/ccao-data/model-sales-val/issues/89) are finished we will manually import the metadata from this run into the new structure. Until then, we can use the commit hash from this PR as a way to save some information.

I will also write metadata parquets locally to be uploaded later.

### DIFF
--- a/manual_flagging/manual_update.py
+++ b/manual_flagging/manual_update.py
@@ -163,31 +163,25 @@ df_condo = pd.merge(df_condo, df_new_groups, on="nbhd", how="left")
 df_condo_to_flag = flg.add_rolling_window(
     df_condo, num_months=inputs["rolling_window_months"]
 )
-# Flag condo outliers, here we remove price per sqft as an input
-# for the isolation forest model since condos don't have a unit sqft
-condo_iso_forest = inputs["iso_forest"].copy()
-condo_iso_forest.remove("sv_price_per_sqft")
-
-condo_stat_groups = ["rolling_window", "geography_split"]
 
 df_condo_flagged = flg_model.go(
     df=df_condo_to_flag,
-    groups=tuple(condo_stat_groups),
-    iso_forest_cols=condo_iso_forest,
+    groups=tuple(inputs["stat_groups"]["condos"]),
+    iso_forest_cols=inputs["iso_forest"],
     dev_bounds=tuple(inputs["dev_bounds"]),
     condos=True,
 )
 
 df_condo_flagged_updated = flg.group_size_adjustment(
     df=df_condo_flagged,
-    stat_groups=condo_stat_groups,
+    stat_groups=inputs["stat_groups"]["condos"],
     min_threshold=inputs["min_groups_threshold"],
     condos=True,
 )
 
 df_condo_flagged_ptax = flg.ptax_adjustment(
     df=df_condo_flagged_updated,
-    groups=condo_stat_groups,
+    groups=inputs["stat_groups"]["condos"],
     ptax_sd=inputs["ptax_sd"],
     condos=True,
 )
--- a/manual_flagging/manual_update.py
+++ b/manual_flagging/manual_update.py
@@ -146,7 +146,7 @@ df["ptax_flag_original"].fillna(False, inplace=True)
 
 # Ingest groups
 df_new_groups = pd.read_excel(
-    os.path.join(root, "QC_salesval_nbhds_condos.xlsx"),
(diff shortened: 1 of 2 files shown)


manual_flagging/manual_update.py
```diff
@@ -152,153 +152,45 @@
 
 # Subset to only City Tri data
 df = df[df["triad_code"] == "1"]
-# Grab only res data
-df_res = df[df["indicator"] == "res"].reset_index(drop=True)
-
-df_res["nbhd"] = df_res["nbhd"].astype(int)
```
jeancochrane, line 158: **[Question, non-blocking]** I see that we're deleting the res update pipeline to make it easier to run the condo pipeline, which makes me kinda nervous. I don't think it's a huge risk since this script is only a stopgap measure anyway, but if we end up needing to make edits to the condo pipeline in the coming days before architecture changes come in (#89), we're going to have to reintroduce this code and remove the condo code and it's going to be a headache. Is there a clean way of using a conditional branch controlled by some type of input flag to tell the script whether to run res vs. condo? Or is the engineering overhead of an approach like that too great to be feasible in the current timeframe?

manual_flagging/yaml/inputs_update.yaml
```diff
@@ -2,19 +2,12 @@
...
-    - "bldg_age_bin"
-    - "char_bldg_sf_bin"
-  multi_family:
-    - "rolling_window"
-    - "geography_split"
-    - "bldg_age_bin"
 
 iso_forest: [
   "meta_sale_price",
-  "sv_price_per_sqft",
```
jeancochrane, line 17: **[Question, non-blocking]** Hmm, is this an indication that the `iso_forest` parameter will need to be different between res and condo groups too? Have we factored that into our redesign (#89)?

manual_flagging/yaml/inputs_update.yaml
```diff
@@ -2,19 +2,12 @@
 # We use these inputs when we want to re-flag already flagged sales.
 ---
 stat_groups:
-  single_family:
+  condos:
```
jeancochrane, line 5: **[Suggestion, non-blocking]** My understanding is that in the future, this data structure will have two top-level keys, `res` and `condo`; I wonder if we could switch to that data structure now so that we can avoid deleting the existing res stat groups?

jeancochrane (approved):
Super messy, but I'm OK doing whatever we need to do in order to give us time to focus on #89.

manual_flagging/yaml/inputs_update.yaml
wagnerlmichael: Yes, I would have had to factor this out. Although we didn't scope it during our meeting. Good catch, I'll incorporate that into the config file.

manual_flagging/manual_update.py
wagnerlmichael: I don't *think* we will need to make changes. However, if we do, I can revert to the other commit that was dealing with the res data. 

A clean way to fix this deletion problem would be to have the architecture changes already completed. I could theoretically build in a bunch of conditionals, but my thinking was that engineering the stop gap wasn't the best use of time since it will mostly be scrapped anyway with the architecture changes. I'm not sure there is a way to partially do it in that we can keep building off of it for the architecture changes. In my head it was simpler to have one commit hash for condos update and one commit hash for res update, but maybe I am thinking about version control in a less-than-ideal way.

manual_flagging/yaml/inputs_update.yaml
wagnerlmichael: Related to this comment - https://github.com/ccao-data/model-sales-val/pull/95#discussion_r1473333219, I could build a temporary config file with boolean switches in the `manual_update.py` to fix this, but I'm guessing it would eventually be removed from main anyways with https://github.com/ccao-data/model-sales-val/issues/89. 

I agree that this structure/way of updating is not very good, but iterating on this on the fly with a time constrain + the time required for our proposed design changes make it so that I don't see a super good way to make an intermediate solution with staying power, happy to talk about this more though.

dfsnow:
@wagnerlmichael Do you need me to take a look at this or are you good to go from Jean's review?

PR #100, opened 2024-02-28
Update stats in readme based on res and condo updates for city tri

wagnerlmichael:
This PR updates our summary statistics on the readme. I made slight changes to the queries. Before, they were based on the `sale.flag` table itself which only worked because all of the sales had only a single flag version. The updated queries make sure that we are looking at the live production flags with the most recent version. To do this we query `default.vw_pin_sale` since it pulls the most updated version of the flag.


jeancochrane (approved):
Good catch here!

dfsnow (approved):

PR #104, opened 2024-03-05
Make flagging script more flexible with respect to geography

### LINKED ISSUE #98: Make flagging script more flexible with respect to geography
Two new requirements that both require a more flexible approach to geography:

1. We would like to be able to run a flagging job against sales within an arbitrary geography or set of geographies (e.g. flag sales within a tri, municipality, or group of neighborhoods)
2. We would like to be able to group sales by arbitrary sets of geographies (e.g. a municipality or a group of neighborhoods)

I am assuming that we don't need to support multiple types of geography per run (e.g. we _do_ need to allow a user to decide between running a flagging job on a tri or a municipality; but we _don't_ need to allow a user to run a flagging job on both a tri _and_ a municipality in the same run).

This will require changes to at least the following parts of the code:

* The query in the flagging script(s)
* The config parameters in `input.yaml`
* The `parameter` output table

Depending on the degree of changes required by the design for this feature, it may be worth sketching out a proposed solution in writing for approval before getting started on the implementation.

wagnerlmichael:
This PR changes the structure and capabilities of the sales-val script. We drew up a new YAML structure based on the assumption that geographies for the statistical groups would be either at the tri, township, or neighborhood level of aggregation. 

This PR also makes a column change to the `sale.parameter` table. We replace the `tri_stat_groups` with `sales_to_run_filter` which allows us to upload a small subset of sales that are calculated using the larger statistical grouping geographies.

Although the groundwork was laid in [this PR](https://github.com/ccao-data/model-sales-val/pull/90), we still have not handled the data migration. I have made a couple changes to the `migrations/0001_update_metadata_structure_after_2023_flag_update.py` file to make the meta data tables more consistent with the new yaml structure. 

Currently we have a backup of the production sales data in a bucket stored as an env variable `AWS_BUCKET_SV` in the migration script. In the migration script, we read from the copied data, manipulate it and write back to a separate directory of that bucket. We also area reading, manipulating , and writing locally saved metadata to this directory. The migration script has been run and the results can be viewed in the DB `z_sales_val_demo`. Specifically I think the parameter table is what requires the most review as that has been changed the most overall relative to current prod structure. When we are satisfied with the new structure, we will remove the prod data and copy this new data into the prod bucket.

Note: env variables representing s3 buckets will need to be changed before merging. Keeping them as they are now pointing to the backup bucket in case changes need to be made.

Closes #98


glue/sales_val_flagging.py
```diff
@@ -219,6 +219,12 @@ def finish_flags(df, start_date, manual_update):
     # Discard pre-2014 data
     df = df[df["meta_sale_date"] >= start_date]
 
+    if sales_to_write_filter["column"]:
+        df = df[
+            df[sales_to_write_filter["column"]].isin(sales_to_write_filter["values"])
+        ]
+        print(df)
+
```
wagnerlmichael, line 227: This is where we implement the `sales_to_write_filter` element from the new yaml structure. If the yaml (yaml object is `inputs`) is empty, it will skip over the logic. If it detects anything it will leverage the column and values keys to make a filter on the data.

glue/sales_val_flagging.py
```diff
@@ -603,7 +609,7 @@ def get_parameter_df(
         "run_filter": [run_filter],
         "iso_forest_cols": [iso_forest_cols],
         "stat_groups": [stat_groups],
-        "tri_stat_groups": [tri_stat_groups],
+        "sales_to_write_filter": [sales_to_write_filter],
```
wagnerlmichael, line 612: Adjusting parameter tables to remove the `tri_stat_groups` and add the `sales_to_write_filter` column. A plus of this is that now we have removed tri from being hard-coded as a column, making the table structure more amenable to future changes.

manual_flagging/flagging.py
```diff
@@ -22,10 +22,11 @@
 
 # Validate the input specification
 # Check housing_market_type
+# TODO: Add res_all and other res_type exclusivity check
 assert "housing_market_type" in inputs, "Missing key: 'housing_market_type'"
 assert set(inputs["housing_market_type"]).issubset(
-    {"res_single_family", "res_multi_family", "condos"}
-), "housing_market_type can only contain 'res_single_family', 'res_multi_family', 'condos'"
+    {"res_single_family", "res_multi_family", "condos", "res_all"}
+), "housing_market_type can only contain 'res_single_family', 'res_multi_family', 'condos', 'res_all'"
```
wagnerlmichael, line 29: Adds new submarket option in the `assert` checks

manual_flagging/flagging.py
```diff
@@ -222,14 +216,15 @@ def create_bins_and_labels(input_list):
...
-
-for tri in current_tris:
+for tri in inputs["run_tri"]:
     # Iterate over housing types defined in yaml
-    for housing_type in inputs["housing_market_type"]:
+    for housing_type in [
+        ht
+        for ht in inputs["housing_market_type"]
+        if ht in inputs["stat_groups"][f"tri{tri}"]
+    ]:
```
wagnerlmichael, line 225: This sub-setting enables a subtle functionality for these yaml inputs:

```yaml
housing_market_type: ["res_single_family", "res_multi_family", "res_all"]
run_tri: [1, 2]
```

This list comprehension makes it so that data frames (on which flagging is performed later in the script) are only created ifn the `housing_market_type` is defined as a subkey under the `tri$n` key in the `stat_groups` yaml object. This is what makes it possible to run flagging over all res properties vs only single family or multi-family.

manual_flagging/flagging.py
```diff
@@ -242,73 +237,33 @@ def create_bins_and_labels(input_list):
...
+                    df_filtered[new_col_name] = pd.cut(
+                        df_filtered[col["column"]], bins=bins, labels=labels
+                    )
+                    transformed_columns.append(new_col_name)
+            else:
+                transformed_columns.append(col)
+
+        dfs_to_rolling_window[key]["columns"] = transformed_columns
+
+        # Add rest of config information
```
wagnerlmichael, line 260: This loop enables take the structure that exists in the `stat_groups` yaml object:
```yaml
      columns:
        - "rolling_window"
        - "geography_split"
        - column: "char_bldg_age"
          bins: [1200, 2400]
        - column: "char_bldg_sf"
          bins: [40]
```
and pre-processes the data such the relevant features are created and correctly processed through the script.

migrations/0001_update_metadata_structure_after_2023_flag_update.py
```diff
@@ -243,15 +248,18 @@ def read_parquets_to_dfs(table):
                 "columns": [
                     "rolling_window",
                     "geography_split",
+                    {"column": "char_bldg_age", "bins": [40]},
+                    {"column": "char_bldg_sf", "bins": [1200, 2400]},
```
wagnerlmichael, line 252: Editing older metadata to be more consistent with current structure.

jeancochrane (approved):
Looking great! My suggestions are almost entirely about adding checks for the input config, but those aren't high priority since we trust that the current config is correct.

I also noticed a few surprises in the output parameter table. Here's what I see in the `z_sales_val_demo.parameter` table in Athena:

| # | run_id | sales_flagged | earliest_data_ingest | latest_data_ingest | run_filter | iso_forest_cols | stat_groups | tri_stat_groups | dev_bounds | ptax_sd | rolling_window | date_floor | short_term_owner_threshold | min_group_thresh |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| 1 | 2024-01-19_18:46-clever-boni | 638958 | 2013-02-01 00:00:00.000 | 2023-12-29 00:00:00.000 | {'housing_market_type': ['condos', 'res_all'], 'run_tri': [1, 2, 3]} | {'res': {'columns': ['meta_sale_price', 'sv_price_per_sqft', 'sv_days_since_last_transaction', 'sv_cgdr', 'sv_sale_dup_counts']}, 'condos': {'columns': ['meta_sale_price', 'sv_days_since_last_transaction', 'sv_cgdr', 'sv_sale_dup_counts']}} | {'og_mansueto': {'res_all': {'columns': ['rolling_window', 'township_code', 'class']}, 'condos': {'columns': ['township_code', 'rolling_window']}}} |  | [2.0, 2.0] | [1.0, 1.0] | 12 | 2014-01-01 | 365 | 30 |
| 2 | 2024-01-29_14:40-pensive-rina | 172239 | 2013-02-01 00:00:00.000 | 2023-12-29 00:00:00.000 | {'housing_market_type': ['res_single_family', 'res_multi_family'], 'run_tri': [1]} | {'res': {'columns': ['meta_sale_price', 'sv_price_per_sqft', 'sv_days_since_last_transaction', 'sv_cgdr', 'sv_sale_dup_counts']}} | {'tri1': {'single_family': {'columns': ['rolling_window', 'geography_split', {'column': 'char_bldg_age', 'bins': [40]}, {'column': 'char_bldg_sf', 'bins': [1200, 2400]}, 'bldg_age_bin', 'char_bldg_sf_bin']}, 'multi_family': {'columns': ['rolling_window', 'geography_split', {'column': 'char_bldg_age', 'bins': [20]}]}}} |  | [2.0, 2.0] | [1.5, 1.5] | 12 | 2014-01-01 | 365 | 30 |
| 3 | 2024-02-01_12:24-nifty-tayun | 117907 | 2013-02-01 00:00:00.000 | 2023-12-29 00:00:00.000 | {'housing_market_type': ['condos'], 'run_tri': [1]} | {'condos': {'columns': ['meta_sale_price', 'sv_days_since_last_transaction', 'sv_cgdr', 'sv_sale_dup_counts']}} | {'condos': {'columns': ['rolling_window', 'geography_split']}} |  | [2.0, 2.0] | [1.5, 1.5] | 12 | 2014-01-01 | 365 | 30 |
: {.responsive}

A few surprises in here for me:
* `tri_stat_groups` is present but null for all runs
* `sales_to_write_filter` is not present
* `run_tri` is not present

Is this expected behavior? Perhaps `z_sales_val_demo` has not been crawled with the latest data?

manual_flagging/flagging.py
```diff
@@ -22,10 +22,11 @@
 
 # Validate the input specification
 # Check housing_market_type
+# TODO: Add res_all and other res_type exclusivity check
```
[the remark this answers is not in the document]
wagnerlmichael: Ah, this TODO is an artifact I forgot to clean up. I believe since we have [this structure](https://github.com/ccao-data/model-sales-val/pull/104#discussion_r1523451996) we don't need to worry about having all 3 res configs at the same time.

Comment from the yaml: 
``` 
# "housing_market_type" and "run_tri" configurations determine the
# specificity of flagging. First, we retrieve the run tris, then run
# the flags on those tris for each selected market type, but only if
# the market type is present in the "stat groups" config under the
# "tri$n" value.
````

File: manual_flagging/flagging.py [hunk 8]
```diff
@@ -246,7 +246,17 @@ def create_bins_and_labels(input_list):
         transformed_columns = []
 
         for col in columns:
-            if isinstance(col, dict) and "column" in col:
+            if isinstance(col, dict):
+                # Validate the structure of the column dictionary
+                # Check if required attributes "column" and "bins" are present
+                # Raise a ValueError if any required attribute is missing
```
jeancochrane, line 252: **[Nitpick, non-blocking]** The implementation that these two comments are describing is pretty straightforward, so I don't think they're necessary. (The `Validate the structure of the column dictionary` comment, on the other hand, is fine by me because it's describing the purpose of the code rather than its implementation.)

```suggestion
```

migrations/0001_update_metadata_structure_after_2023_flag_update.py
```diff
@@ -110,7 +110,7 @@ def read_parquets_to_dfs(table):
 dfs_sale_parameter["2024-01-19_18:46-clever-boni"]["stat_groups"] = str(
     {
         "og_mansueto": {
-            "res": {"columns": ["rolling_window", "township_code", "class"]},
+            "res_all": {"columns": ["rolling_window", "township_code", "class"]},
```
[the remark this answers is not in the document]
jeancochrane: Looks good to me! I don't think we need to be strict about preserving the legacy structure, as long as we retain all of the relevant information about the parameters used in the run.

manual_flagging/flagging.py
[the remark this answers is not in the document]
jeancochrane: I have a quick question about this, but I'll ask IRL!

jeancochrane (approved):
Good to go! One tiny nitpick, but merge away once you've resolved it to your liking.

PR #107, opened 2024-03-13
Pull neighborhood groups from a table in the data warehouse

jeancochrane:
This PR refactors `manual_flagging/flagging.py` to pull neighborhood group definitions from the data warehouse instead of a local Excel workbook.

This is a companion PR to https://github.com/ccao-data/data-architecture/pull/343 and should only be merged once that PR has been deployed.


manual_flagging/flagging.py
```diff
@@ -57,6 +57,15 @@
         BETWEEN DATE '{date_floor}'
         AND DATE '{inputs['time_frame']['end']}')"""
 
+# Parse run_date and make sure its format is correct
+date_fmt = "%Y-%m-%d"
+run_date = (
```
jeancochrane, line 62: I added a note on what's going on with `run_date` to its definition in the `inputs.yml` file below.

data/res_condos_nbhd_groups_2024.xlsx
jeancochrane, line 1: This file is no longer necessary since the groups will be present in the data warehouse once https://github.com/ccao-data/data-architecture/pull/343 lands.

jeancochrane, line 1: This file is no longer necessary since the groups will be present in the data warehouse once https://github.com/ccao-data/data-architecture/pull/343 lands.

manual_flagging/flagging.py
```diff
@@ -94,6 +94,19 @@
...
+
+-- Select neighborhood groups and filter for most recent versions
+neighborhood_group AS (
+    SELECT nbhd_group.nbhd, nbhd_group.group_name
+    FROM location.neighborhood_group AS nbhd_group
+    INNER JOIN (
+        SELECT nbhd, MAX(version) AS version
+        FROM location.neighborhood_group
+        GROUP BY nbhd
+    ) AS latest_group_version
```
jeancochrane, line 107: Note that this join could also be extended to identify the version of the neighborhood groups that was used for a specific run; see the removed code in 70af4829de14313652ead423b5dc6c007da53aca for a sketch of what that would look like.

wagnerlmichael (approved):
This looks good to me! I also read through https://github.com/ccao-data/data-architecture/pull/343/files and I like how it is open-ended for future development scenarios.

PR #108, opened 2024-03-13
Add  run_note to metadata and improve date range param data

wagnerlmichael:
In this PR we add a run note to the metadata tables and change `date_floor` to `date_range` in the parameter table. We also edit the migration script to account for these changes.


glue/sales_val_flagging.py
```diff
@@ -554,6 +554,7 @@ def to_float(value):
         "run_filter",
         "iso_forest_cols",
         "stat_groups",
+        "date_range",
```
jeancochrane, line 557: **[Thought, non-blocking]** It might be nice for this param to be called `time_frame` so that it matches the value in `inputs.yml` exactly, but I don't feel super strongly about it -- `date_range` is clear in spite of being slightly different!

jeancochrane (approved):
Thanks for the quick work on this!

glue/sales_val_flagging.py
wagnerlmichael: Good call, fixed.

jeancochrane (approved):
Good to go, thanks for this!

Issue #80, opened 2023-12-04
Update CI prefix to match dbt


dfsnow:
https://github.com/ccao-data/data-architecture/pull/259 switched the CI database schema prefix to `z_ci_` for better sorting. We should do the same in the Terraform config files in this repository. Low priority.

PR #109, opened 2024-03-14
Change `ci_*` prefix to `z_ci_*` in Terraform resources

jeancochrane:
This PR updates the prefixes used by the resources defined in the Terraform config to match the convention we use in https://github.com/ccao-data/data-architecture, namely that staging resources get prefixed with `z_ci_*` instead of `ci_*` for better sorting in the AWS UI.

See [here](https://github.com/ccao-data/model-sales-val/actions/runs/8286236192/job/22675921607?pr=109) for an example of a workflow run that successfully sets up resources with the new prefix, and [here](https://github.com/ccao-data/model-sales-val/actions/runs/8286446841/job/22676451090) for an example of an equivalent `cleanup-terraform` run.

Closes https://github.com/ccao-data/model-sales-val/issues/80.

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,7 +314,7 @@ This repository manages the configurations, scripts, and details for an AWS Glue
 2. All changes to these components should originate from this repository. This ensures that every modification is tracked and version-controlled.
 3. The **only** advisable actions in the AWS Console concerning this Glue job are:
     - Running the job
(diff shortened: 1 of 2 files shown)


jeancochrane:
Closing to test the cleanup step.

dfsnow (approved):
Nice. The perfectionist in me has been waiting for this one.

PR #110, opened 2024-03-15
Refresh readme

wagnerlmichael:
This PR updates the athena table diagrams and the proportion of outliers based on recent runs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,13 +296,13 @@ erDiagram
     }
 
     group_mean {
-        string group
+        string group PK
         double group_mean
         double group_std
         double group_sqft_std
         double group_sqft_mean
         bigint group_size
-        string run_id
+        string run_id PK
     }
 ```
 
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ ORDER BY
 
 |Outlier Type           |Percentage|
 |-----------------------|---------:|
-|PTAX-203 flag (Low)    |35.5%     |
+|PTAX-203 flag (Low)    |35.50%     |
 |Non-person sale (low)  |18.17%    |
 |Non-person sale (high) |10.29%    |
 |Anomaly (high)         |7.08%     |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -116,7 +116,7 @@ FROM
     TotalRecords, OutlierCount;
 -->
 
-As of 2024-02-28, around **6.9%** of the total sales have some sort of outlier classification. Within that 6.9%, the proportion of each outlier type is:
+As of 2024-03-15, around **6.9%** of the total sales have some sort of outlier classification. Within that 6.9%, the proportion of each outlier type is:
```
jeancochrane, line 119: **[Question, non-blocking]** Is the 6.9% number still correct given the changes to the percentages in the table below?

README.md
```diff
@@ -148,24 +148,24 @@ ORDER BY
...
-|Low price (raw & sqft) |4.09%     |
-|PTAX-203 flag (High)   |3.07%     |
-|Home flip sale (high)  |2.26%     |
-|High price (sqft)      |2.03%     |
-|Low price (sqft)       |1.99%     |
-|Anomaly (low)          |1.65%     |
-|High price (raw & sqft)|1.63%     |
-|Home flip sale (low)   |1.15%     |
-|Family sale (low)      |0.65%     |
+|PTAX-203 flag (Low)    |35.5%     |
```
jeancochrane, line 151: **[Nitpick, non-blocking]** One small readability suggestion for this table: If we were to use two decimal places for every percentage instead of truncating the multiples of 10 to one decimal place, then the order of magnitude of each percentage would be easy to gauge just by giving the table a quick scan since each percentage's decimal point would align vertically.

Not a necessary change, since I know the truncation pattern predated this PR, but it's an easy way we could improve the readability of the table if that interests you!

```suggestion
|PTAX-203 flag (Low)    |35.50%     |
```

jeancochrane (approved):
Looks great, thanks!

README.md
wagnerlmichael: Yep!

README.md
wagnerlmichael: Good eye!

PR #111, opened 2024-03-15
Small readability tweak

wagnerlmichael:
Standardize decimal places in readme table showing outlier type makeup.


jeancochrane (approved):
Thanks for the quick fix!

PR #112, opened 2024-03-26
Test centralized GitHub actions config

wagnerlmichael:
Instantiate PR

### DIFF
--- a/.github/workflows/pre-commit.yaml
+++ b/.github/workflows/pre-commit.yaml
@@ -1,4 +1,3 @@
----
 on:
   pull_request:
   push:
@@ -10,24 +9,9 @@ jobs:
   pre-commit:
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout
-        uses: actions/checkout@v2
-
-      - name: Set up Python
-        uses: actions/setup-python@v2
+      - name: Pre-Commit Action
+        uses: |
+          ccao-data/common-actions/.github/actions/pre-commit\
+          @create-composite-pre-commit-action
         with:
-          python-version: 3.10.12
-
-      - name: Install pre-commit
-        run: sudo apt-get install pre-commit
-        shell: bash
-
-      - name: Cache pre-commit environment
-        uses: actions/cache@v2
-        with:
-          path: ~/.cache/pre-commit
-          key: pre-commit-${{ hashFiles('.pre-commit-config.yaml') }}
-
-      - name: Run pre-commit
-        run: pre-commit run --show-diff-on-failure --color=always --all-files
-        shell: bash
+          python-version: '3.10.12'
--- a/manual_flagging/flagging.py
+++ b/manual_flagging/flagging.py
@@ -16,6 +16,8 @@ from pyathena.pandas.util import as_pandas
 root = sp.getoutput("git rev-parse --show-toplevel")
 os.chdir(os.path.join(root, "manual_flagging"))
 
+print("----------------------------------------------------------------------------------")
+
 # Use yaml as inputs
 with open(os.path.join("yaml", "inputs.yaml"), "r") as stream:
     inputs = yaml.safe_load(stream)


.github/workflows/pre-commit.yaml
```diff
@@ -10,24 +9,11 @@ jobs:
...
-      - name: Cache pre-commit environment
-        uses: actions/cache@v2
-        with:
-          path: ~/.cache/pre-commit
-          key: pre-commit-${{ hashFiles('.pre-commit-config.yaml') }}
-
-      - name: Run pre-commit
-        run: pre-commit run --show-diff-on-failure --color=always --all-files
-        shell: bash
+          hooks: "shellcheck yamllint black"
```
wagnerlmichael, line 18: Space separated list to be parsed in bash script in composite repo.

.gitignore
```diff
@@ -2,6 +2,7 @@ venv/
 __pycache__/
 ipynb_checkpoints
 glue/schema/*.parquet
+.pre-commit-config.yaml
```
wagnerlmichael, line 5: Add this local version of the `.pre-commit-config.yaml` to the git ignore. This allows the use to have control over their local hooks. Upon PR or push to main/master is when the factored out actions would take effect.

.github/workflows/pre-commit.yaml
wagnerlmichael, line 18: This is where the hooks are specified

.github/workflows/pre-commit.yaml
```diff
@@ -10,24 +9,11 @@ jobs:
   pre-commit:
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout
+      - name: Checkout main code
         uses: actions/checkout@v2
 
-      - name: Set up Python
-        uses: actions/setup-python@v2
+      - name: Pre-Commit Action
```
jeancochrane, line 15: **[Nitpick, non-blocking]** I would call this something slightly more specific, like:

```suggestion
      - name: Run pre-commit checks
```

.github/workflows/pre-commit.yaml
```diff
@@ -10,24 +9,11 @@ jobs:
...
-        uses: actions/cache@v2
-        with:
-          path: ~/.cache/pre-commit
-          key: pre-commit-${{ hashFiles('.pre-commit-config.yaml') }}
-
-      - name: Run pre-commit
-        run: pre-commit run --show-diff-on-failure --color=always --all-files
-        shell: bash
+          hooks: "shellcheck yamllint black"
+          python-version: '3.10.12'
```
jeancochrane, line 19: **[Nitpick, non-blocking]** I think this isn't necessary since it's not used by [the action](https://github.com/ccao-data/actions/pull/22/files#diff-5769c037baef7e92c33706e693ad9cdb4ae31f2632659aa1ec0f0544979b8f0e) anymore, right?

```suggestion
```

.gitignore
```diff
@@ -2,8 +2,10 @@ venv/
...
```
jeancochrane, line 5: **[Thought, non-blocking]** After thinking about it for a bit, I think that I actually prefer the other side of the tradeoff described in the PR body: It strikes me as slightly friendlier to keep a local copy of the pre-commit config under version control, in spite of the duplication with the `ccao-data/actions/pre-commit` action, so that developers can run pre-commit locally with minimal overhead. Having to occasionally resolve conflicts between the local pre-commit config and the remote config in `ccao-data/actions/pre-commit` will sometimes be annoying, but I expect it will be a less common annoyance than requiring every developer to copy the remote config to their local repo before they start development (and then also update their local copy whenever the remote changes).

That being said, this is a weakly-held opinion; what do you two think?

.gitignore
```diff
@@ -2,8 +2,10 @@ venv/
...
 *.csv
 
+
```
jeancochrane, line 8: **[Nitpick, non-blocking]** Stray line break here:

```suggestion
```

jeancochrane (approved):
I really like this interface, and I think it's 95% there, with a remaining open question around the way that local pre-commit should be handled. I think this could come in immediately if need be, but am curious to hear what Dan thinks about the local pre-commit issue first!

.gitignore
wagnerlmichael: I agree with the sentiment. A question for me is: Given that we would include both in each repository, how valuable is it to transition to composite actions across all of our repositories? It doesn't seem clear to me that we would be significantly reducing lines of code or reducing the total number of files. 

What would positives from this be? I suppose we can have a single source of truth for which to look for all of our hook versions in the actions repo, and that at least upon PRs they are all standardized. Maybe this would be a good start for other actions functionalities I'm not aware of.

jeancochrane: Good point, there's definitely a tradeoff between DRY and developer experience here. I'm comfortable moving forward with the gitignore and just making sure we explain to the rest of the team how they'll be expected to manage local pre-commit configs going forward.

dfsnow: See my response on the [actions PR](https://github.com/ccao-data/actions/pull/22#pullrequestreview-2006915791)

dfsnow (changes requested):
See corresponding Actions PR

.github/workflows/pre-commit.yaml
```diff
@@ -10,24 +9,8 @@ jobs:
...
-        uses: actions/cache@v2
-        with:
-          path: ~/.cache/pre-commit
-          key: pre-commit-${{ hashFiles('.pre-commit-config.yaml') }}
-
-      - name: Run pre-commit
-        run: pre-commit run --show-diff-on-failure --color=always --all-files
-        shell: bash
+      - name: Run pre-commit checks
+        uses: ccao-data/actions/pre-commit@create-composite-pre-commit-action
```
wagnerlmichael, line 16: Eventually this branch name will be subbed for main when we merge.

jeancochrane (approved):
Awesome 👏🏻

dfsnow (approved):
👍

PR #114, opened 2024-03-27
Add `scripts/export.py` script for exporting flags to iasWorld

### LINKED ISSUE #113: Export flags to CSV for iasWorld upload
Task 1 in https://github.com/ccao-data/model-sales-val/issues/102.

jeancochrane:
This PR adds a new script to the repo allowing a caller to export flags to iasWorld for upload.

## Testing

To test, check out the code and run the script:

```
python3 -m venv venv
source venv/bin/activate
pip install -r manual_flagging/requirements.txt
python3 scripts/export.py > sales_val_flags.csv
```

You may have to run `aws-mfa` if you haven't already today.

## Questions

A couple of big-picture design questions about this code:

* Should this script live here in `model-sales-val`, or instead in https://github.com/ccao-data/data-architecture? On the one hand, it largely references data produced by this repo, but on the other hand it represents a type of export that is currently uncommon but that we expect to become more common across our data warehouse in the future. From a code perspective, there's no reason why it needs to live in one repo or the other, but once we start doing exports like this more regularly I wonder if we'll want them all to live in one place.
* What sorts of flags should the script accept for filtering the flag output? And should we implement those now, or wait until we know more about our export cadence? I think it makes sense to wait, but I'm curious what you think in general. Some ideas:
    * Filter by one or more run IDs, so that e.g. we can generate a new run and then only export the flags it created
    * Filter by run date, so we can e.g. export all flags created after a certain date 
    * Any others?

Closes #113.

### DIFF
--- a/scripts/export.py
+++ b/scripts/export.py
@@ -19,10 +19,9 @@ SALE_KEY_FIELD = "SALEKEY"
 IS_OUTLIER_FIELD = "USER26"
 OUTLIER_TYPE_FIELD = "USER27"
 RUN_ID_FIELD = "USER28"
-ANALYST_DETERMINATION_FIELD = "USER29"
-ANALYST_REVIEW_DATE_FIELD = "UDATE1"
 
 OUTLIER_TYPE_CODES = {
+    "Not outlier": "0",
     "Anomaly (high)": "1",
     "Anomaly (low)": "2",
     "Family sale (high)": "3",
@@ -40,8 +39,7 @@ OUTLIER_TYPE_CODES = {
     "Non-person sale (high)": "15",
     "Non-person sale (low)": "16",
     "PTAX-203 flag (High)": "17",
-    "PTAX-203 flag (Low)": "17",
-    "Not outlier": pandas.NA,
+    "PTAX-203 flag (Low)": "18",
 }
 
 if __name__ == "__main__":
@@ -84,9 +82,7 @@ if __name__ == "__main__":
             ELSE 'N'
         END AS {IS_OUTLIER_FIELD},
         flag.sv_outlier_type AS {OUTLIER_TYPE_FIELD},
-        flag.run_id AS {RUN_ID_FIELD},
-        'N' AS {ANALYST_DETERMINATION_FIELD},
-        NULL AS {ANALYST_REVIEW_DATE_FIELD}
+        flag.run_id AS {RUN_ID_FIELD}
     FROM sale.flag AS flag
     -- Filter flags for the most recent version
     INNER JOIN ({FLAG_LATEST_VERSION_QUERY}) AS flag_latest_version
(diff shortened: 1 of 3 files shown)


scripts/export.py
```diff
@@ -0,0 +1,139 @@
...
+    "Home flip sale (low)": "10",
+    "Low price (raw & sqft)": "11",
+    "Low price (raw)": "12",
+    "Low price (sqft)": "13",
+    "Low price swing": "14",
+    "Non-person sale (high)": "15",
+    "Non-person sale (low)": "16",
+    "PTAX-203 flag (High)": "17",
+    "PTAX-203 flag (Low)": "17",
+    "Not outlier": pandas.NA,
```
jeancochrane, line 44: They seem reasonable to me, but I'm not 100% sure of these two choices, and we need to confirm them with Mirella:

* High and low PTAX-203 flags should share the same code, since there is only one `PTAX-203 flag` code (17)
* Sales that have been flagged as _not_ being outliers should be uploaded to iasWorld with `USER26 = 'N'` (not an outlier) and `USER27 = ""` (no reason code)

scripts/export.py
```diff
@@ -0,0 +1,139 @@
...
+}
+
+if __name__ == "__main__":
+    # Setup a logger that logs to stderr so it does not get captured as part
+    # of the script's data output
+    logger = logging.getLogger()
+    logger.setLevel(logging.INFO)
+    logger.addHandler(logging.StreamHandler(sys.stderr))
```
jeancochrane, line 52: This is a little bit janky, and all comes back to the fact that I prefer scripts that output data to stdout so that you can redirect or pipe their output wherever you like. I'm willing to refactor things to log to stdout and save results to a preset file instead if you feel strongly about it.

scripts/export.py
```diff
@@ -0,0 +1,139 @@
...
+        (flag_df[IS_OUTLIER_FIELD] == "Y") & (flag_df[OUTLIER_TYPE_FIELD].isna())
+    ].empty, f"{OUTLIER_TYPE_FIELD} cannot be null when {IS_OUTLIER_FIELD} is Y"
+
+    assert flag_df[
+        (flag_df[IS_OUTLIER_FIELD] == "N") & (~flag_df[OUTLIER_TYPE_FIELD].isna())
+    ].empty, f"{OUTLIER_TYPE_FIELD} must be null when {IS_OUTLIER_FIELD} is N"
+
+    assert (
+        num_flags == expected_num_flags
+    ), f"Expected {expected_num_flags} flagged sales, got {num_flags}"
```
jeancochrane, line 135: While normally it would be more appropriate to implement checks like these in our dbt tests, my thinking is that we want some guardrails that will warn us if we modify the flag query and accidentally introduce a bad join or select that clobbers the data in some unanticipated way.

scripts/export.py
```diff
@@ -0,0 +1,143 @@
...
+        (flag_df[IS_OUTLIER_FIELD] == "N")
+        & (flag_df[OUTLIER_TYPE_FIELD] != OUTLIER_TYPE_CODES["Not outlier"])
+    ].empty, (
+        f"{OUTLIER_TYPE_FIELD} must be {OUTLIER_TYPE_CODES['Not outlier']} "
+        f"when {IS_OUTLIER_FIELD} is N"
+    )
+
+    assert (
+        num_flags == expected_num_flags
+    ), f"Expected {expected_num_flags} total sales, got {num_flags}"
```
jeancochrane, line 139: At some point it would be nice to copy these tests over to dbt to test the iasWorld flag table, but however we decide to do that, I still think it's useful to have some guardrails here that will warn us if we modify the flag query and accidentally introduce a bad join or select that clobbers the data in some unanticipated way.

dfsnow: **praise:** Agree we should move it over, but this is great. Would like to see more defensive coding like this elsewhere in our repos.

dfsnow (approved):
@jeancochrane This is great! Thanks for the quick turnaround. To answer your questions:

> Should this script live here in model-sales-val, or instead in https://github.com/ccao-data/data-architecture? On the one hand, it largely references data produced by this repo, but on the other hand it represents a type of export that is currently uncommon but that we expect to become more common across our data warehouse in the future. From a code perspective, there's no reason why it needs to live in one repo or the other, but once we start doing exports like this more regularly I wonder if we'll want them all to live in one place.

IMO it should probably live `data-architecture` long-term. I imagine most of our "moving data around" scripts will live there.

> What sorts of flags should the script accept for filtering the flag output? And should we implement those now, or wait until we know more about our export cadence? I think it makes sense to wait, but I'm curious what you think in general. Some ideas:
Filter by one or more run IDs, so that e.g. we can generate a new run and then only export the flags it created
Filter by run date, so we can e.g. export all flags created after a certain date
Any others?

The main one I expect will be something like "export flags for sales without a flag in iasWorld" or "export flags that overwrite this existing run". But we can wait on the implementation until we know a bit more after next week.

wagnerlmichael (approved):
Looks good to me!

jeancochrane:
> IMO it should probably live data-architecture long-term. I imagine most of our "moving data around" scripts will live there.

That makes sense to me @dfsnow -- I'll merge the script here so we can start using it immediately and then open up an issue to move it to `data-architecture`.

PR #121, opened 2024-04-24
Add intern names to generator

dfsnow:
Adds recent interns to the name generator for the sales val model. I don't love that this list is now shared across the `ccao` package and this model. I wonder if there's a way we could share it. Maybe use a git submodules. Any thoughts @wagnerlmichael @jeancochrane?

### DIFF
--- a/glue/sales_val_flagging.py
+++ b/glue/sales_val_flagging.py
@@ -427,6 +427,10 @@ def finish_flags(df, start_date, manual_update, sales_to_write_filter):
         "jean",
         "damon",
         "michael",
+        "matt",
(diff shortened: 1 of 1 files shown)


jeancochrane (approved):
Looks good! By my estimation, the root problem is that we publish the `ccao` package for R but there's no equivalent for Python, which I expect will continue to cause problems if we keep writing Python code. Perhaps we could start publishing one incrementally by restricting the scope to just the run ID name generator at first. But if we feel like that's too much of a lift, we could also just store this list as a seed in the warehouse and call it a day,

PR #123, opened 2024-05-16
Refactor run_id generation

### LINKED ISSUE #122: Factor out run_id generation
We need to factor out the `run_id` generation by pulling it in from a new dbt seed https://github.com/ccao-data/data-architecture/pull/451#issue-2298593802

wagnerlmichael:
We take advantage of the dbt seed name generation tables to refactor `run_id` generation, reducing LoC and a package dependency.

Closes #122.


dfsnow (approved):
Works for me!

Issue #125, opened 2024-05-23
Rename reason codes to match the new spec


jeancochrane:
The [updated sales val design doc](https://cookcounty-my.sharepoint.com/:w:/r/personal/jean_cochrane_cookcountyil_gov/Documents/Design%20Doc%20-%20Updated%20Sales%20Validation%20Specification.docx?d=wd6a2a31a0a664b168522078b16b3f92a&csf=1&web=1&e=r32MGj&nav=eyJoIjoiMTE0ODg1MzU5MyJ9) includes some new names for our flags that should be easier for analysts to use.

Following up on https://github.com/ccao-data/model-sales-val/issues/124, we need to change our reason codes to match the new spec. Here's how codes should change:

* `Home flip sale` -> `Short-term owner`
* `Anomaly` -> `Statistical anomaly`
* `Price swing` -> `Price swing / home flip`
* `PTAX-203 flag` -> `PTAX-203 exclusion`

Issue #124, opened 2024-05-23
Refactor the pipeline to output multiple reason codes


jeancochrane:
Currently, the sales val pipeline outputs reason codes for each flag that combine a characteristic reason with a price reason. For example, `Family sale (high)` is a combination of a characteristic reason (family sale) with a price reason (high price).

Per the [updated sales val design doc](https://cookcounty-my.sharepoint.com/:w:/r/personal/jean_cochrane_cookcountyil_gov/Documents/Design%20Doc%20-%20Updated%20Sales%20Validation%20Specification.docx?d=wd6a2a31a0a664b168522078b16b3f92a&csf=1&web=1&e=PN3AaT&nav=eyJoIjoiMTc4ODE0MDM3MiJ9), we need to refactor the pipeline so that it outputs multiple reason codes for each flag. This requires us to disaggregate our characteristic reasons from our price reasons, and update the pipeline to output one characteristic reason and one price reason for each flagged sale.

This is what the refactored set of reason codes should look like:

* Characteristic reasons
  * `Short-term owner`
  * `PTAX-203 Exclusion`
  * `Family sale`
  * `Non-person sale`
  * `Statistical Anomaly`
  * `Price swing / Home Flip`
* Price reasons
  * `High price`
  * `High price per square foot`
  * `Low price`
  * `Low price per square foot`

PR #128, opened 2024-05-31
Implement design doc changes

wagnerlmichael:
This PR handles https://github.com/ccao-data/model-sales-val/issues/125 and https://github.com/ccao-data/model-sales-val/issues/124.

Essentially I take the existing system and make changes such that it fits with these issues and the new design doc. 

The biggest change here is that we remove `sv_outlier_type` and introduce 3 columns:
- `sv_outlier_reason1`
- `sv_outlier_reason2`
- `sv_outlier_reason3`

In the existing set up, if we have an outlier that is classified , then `sv_outlier_type()` must be one of the following:
- A price outlier defined by the `dev_bounds: [2, 2]` definition in `params.yaml`
- Above price outlier + a characteristic reason determined in the `outlier_type()` function in `glue/flagging_script_glue/flagging.py`. 
- One of the ptax questions flagged and a price deviation from `ptax_sd: [1, 1]`defined in `params.yaml`

The following changes are implemented in order to maintain the existing specs and bring in the new structure. In `sv_outlier_type()` the price and characteristic outlier types are split into `char_conditions`/`char_labels` and `price_conditions`/`price_labels`. 

For the three `sv_outlier_reason$n` columns, we populate them with the following order of priority

1. PTAX-203 Exclusion
2. Price label
3. Char label

Then, we assign the observation as an outlier only if `sv_is_outlier1` is a price label or a ptax label. This allows us to retain sales validation information even if we don't want to technically classify the sale as an outlier. Here are some scenarios:

Scenario 1:
- Outlier  Classification: True
- `sv_outlier_reason1` - High Price
- `sv_outlier_reason2` - High Price (sqft)
- `sv_outlier_reason3` - Non-person sale

Scenario 2:
- Outlier  Classification: False
- `sv_outlier_reason1` - Non-person sale
- `sv_outlier_reason2` - Statistical anomaly
- `sv_outlier_reason3` - null

### DIFF
--- a/glue/flagging_script_glue/flagging.py
+++ b/glue/flagging_script_glue/flagging.py
@@ -742,9 +742,8 @@ def z_normalize_groupby(s: pd.Series):
 
 def outlier_type(df: pd.DataFrame, condos: bool) -> pd.DataFrame:
     """
-    Enhances the dataframe by adding two columns for outlier reasons:
-    - sv_char_outlier_reason: Reasons based on characteristics including price swings.
-    - sv_char_price_reason: Reasons based on high or low pricing levels.
+    This function create indicator columns for each distinct outlier type between price
+    and characteristic outliers. These columns are prefixed with 'sv_ind_'.
 
     Inputs:
         df (pd.DataFrame): Dataframe with necessary columns created from previous functions.
--- a/glue/flagging_script_glue/flagging.py
+++ b/glue/flagging_script_glue/flagging.py
@@ -828,8 +828,6 @@ def outlier_type(df: pd.DataFrame, condos: bool) -> pd.DataFrame:
             .eq("legal_entity")
             .any(axis=1),
             df["sv_anomaly"] == "Outlier",
-            df["sv_pricing"].str.contains("High price swing"),
-            df["sv_anomaly"] == "Outlier",
             df["sv_pricing"].str.contains("High price swing")
             | df["sv_pricing"].str.contains("Low price swing"),
         ]
--- a/glue/flagging_script_glue/flagging.py
+++ b/glue/flagging_script_glue/flagging.py
@@ -785,21 +785,22 @@ def outlier_type(df: pd.DataFrame, condos: bool) -> pd.DataFrame:
     else:
         # Define conditions for price-based reasons
         price_conditions = [
+            (
+                df["sv_pricing"].str.contains("High")
+                & (df["sv_which_price"].str.contains("raw"))
+            ),
+            (
+                df["sv_pricing"].str.contains("Low")
+                & (df["sv_which_price"].str.contains("raw"))
+            ),
             (df["sv_pricing"].str.contains("High"))
-            & (df["sv_which_price"] == "(raw & sqft)"),
+            & (df["sv_which_price"].str.contains("sqft")),
             (df["sv_pricing"].str.contains("Low"))
-            & (df["sv_which_price"] == "(raw & sqft)"),
-            (df["sv_pricing"].str.contains("High") & (df["sv_which_price"] == "(raw)")),
-            (df["sv_pricing"].str.contains("Low") & (df["sv_which_price"] == "(raw)")),
-            (df["sv_pricing"].str.contains("High"))
-            & (df["sv_which_price"] == "(sqft)"),
-            (df["sv_pricing"].str.contains("Low")) & (df["sv_which_price"] == "(sqft)"),
+            & (df["sv_which_price"].str.contains("sqft")),
         ]
 
         # Define labels for price-based reasons
         price_labels = [
-            "sv_ind_price_high_price_raw_and_sqft",
-            "sv_ind_price_low_price_raw_and_sqft",
             "sv_ind_price_high_price",
             "sv_ind_price_low_price",
             "sv_ind_price_high_price_sqft",
@@ -814,25 +815,6 @@ def outlier_type(df: pd.DataFrame, condos: bool) -> pd.DataFrame:
     # Create indicator columns for each flag type
     for label, condition in outlier_type_dict.items():
         df[label] = condition.astype(int)
-
-    if not condos:
-        # Separate out combined raw and sqft label to be discrete cols
-        df.loc[
-            df["sv_ind_price_high_price_raw_and_sqft"] == 1,
-            ["sv_ind_price_high_price_sqft", "sv_ind_price_high_price"],
-        ] = 1
-        df.loc[
-            df["sv_ind_price_low_price_raw_and_sqft"] == 1,
-            ["sv_ind_price_low_price_sqft", "sv_ind_price_low_price"],
-        ] = 1
-
-        df.drop(
-            columns=[
-                "sv_ind_price_high_price_raw_and_sqft",
-                "sv_ind_price_low_price_raw_and_sqft",
-            ],
-            inplace=True,
-        )
 
     return df
 
--- a/glue/flagging_script_glue/flagging.py
+++ b/glue/flagging_script_glue/flagging.py
@@ -786,8 +786,8 @@ def outlier_type(df: pd.DataFrame, condos: bool) -> pd.DataFrame:
             .eq("legal_entity")
             .any(axis=1),
             df["sv_anomaly"] == "Outlier",
-            df["sv_pricing"].str.contains("High price swing"),
-            df["sv_pricing"].str.contains("Low price swing"),
+            df["sv_pricing"].str.contains("High price swing")
+            | df["sv_pricing"].str.contains("Low price swing"),
         ]
 
         # Define labels for characteristic-based reasons
@@ -796,8 +796,7 @@ def outlier_type(df: pd.DataFrame, condos: bool) -> pd.DataFrame:
             "Family sale",
             "Non-person sale",
             "Statistical anomaly",
-            "High price swing",
-            "Low price swing",
+            "Price Swing / Home flip",
         ]
 
         # Define conditions for price-based reasons
@@ -830,7 +829,9 @@ def outlier_type(df: pd.DataFrame, condos: bool) -> pd.DataFrame:
             .any(axis=1),
             df["sv_anomaly"] == "Outlier",
             df["sv_pricing"].str.contains("High price swing"),
-            df["sv_pricing"].str.contains("Low price swing"),
+            df["sv_anomaly"] == "Outlier",
+            df["sv_pricing"].str.contains("High price swing")
+            | df["sv_pricing"].str.contains("Low price swing"),
         ]
 
         # Define labels for characteristic-based reasons
@@ -839,8 +840,7 @@ def outlier_type(df: pd.DataFrame, condos: bool) -> pd.DataFrame:
             "Family sale",
             "Non-person sale",
             "Statistical anomaly",
-            "High price swing",
-            "Low price swing",
+            "Price Swing / Home flip",
         ]
 
         # Define conditions for price-based reasons
--- a/glue/flagging_script_glue/flagging.py
+++ b/glue/flagging_script_glue/flagging.py
@@ -818,23 +818,24 @@ def outlier_type(df: pd.DataFrame, condos: bool) -> pd.DataFrame:
     for label, condition in outlier_type_dict.items():
         df[label] = condition.astype(int)
 
-    # Separate out combined raw and sqft label to be discrete cols
-    df.loc[
-        df["sv_ind_price_high_price_raw_and_sqft"] == 1,
-        ["sv_ind_price_high_price_sqft", "sv_ind_price_high_price"],
-    ] = 1
-    df.loc[
-        df["sv_ind_price_low_price_raw_and_sqft"] == 1,
-        ["sv_ind_price_low_price_sqft", "sv_ind_price_low_price"],
-    ] = 1
+    if not condos:
+        # Separate out combined raw and sqft label to be discrete cols
+        df.loc[
+            df["sv_ind_price_high_price_raw_and_sqft"] == 1,
+            ["sv_ind_price_high_price_sqft", "sv_ind_price_high_price"],
+        ] = 1
+        df.loc[
+            df["sv_ind_price_low_price_raw_and_sqft"] == 1,
+            ["sv_ind_price_low_price_sqft", "sv_ind_price_low_price"],
+        ] = 1
 
-    df.drop(
-        columns=[
-            "sv_ind_price_high_price_raw_and_sqft",
-            "sv_ind_price_low_price_raw_and_sqft",
-        ],
-        inplace=True,
-    )
+        df.drop(
+            columns=[
+                "sv_ind_price_high_price_raw_and_sqft",
+                "sv_ind_price_low_price_raw_and_sqft",
+            ],
+            inplace=True,
+        )
 
     return df
 
--- a/glue/flagging_script_glue/flagging.py
+++ b/glue/flagging_script_glue/flagging.py
@@ -83,12 +83,8 @@ def outlier_taxonomy(df: pd.DataFrame, permut: tuple, groups: tuple, condos: boo
     """
 
     df = check_days(df, SHORT_TERM_OWNER_THRESHOLD)
-
     df = pricing_info(df, permut, groups, condos=condos)
-
     df = outlier_type(df, condos=condos)
-    # df = outlier_flag(df)
-    # df = special_flag(df)
 
     return df
 
@@ -266,26 +262,6 @@ def pricing_info(
 
     if not condos:
         df["sv_which_price"] = df.apply(which_price, args=(holds, groups), axis=1)
-
-    return df
-
-
-def special_flag(df: pd.DataFrame) -> pd.DataFrame:
-    """
-    Creates column that checks whether there is a special flag for this record.
-    Inputs:
-        df (pd.DataFrame): dataframe to add flags onto
-    Outputs:
-        df (pd.DataFrame): dataframe with 'special_flags' column
-    """
-    cond = [
-        (df["sv_name_match"] != "No match"),
-        (df["sv_short_owner"] == "Short-term owner"),
-        (df["sv_transaction_type"] == "legal_entity-legal_entity"),
-    ]
-    labels = ["Family sale", "Home flip sale", "Non-person sale"]
-
-    df["sv_special_flags"] = np.select(cond, labels, default="Not special")
 
     return df
 
--- a/glue/flagging_script_glue/flagging.py
+++ b/glue/flagging_script_glue/flagging.py
@@ -767,114 +767,107 @@ def z_normalize_groupby(s: pd.Series):
 
 def outlier_type(df: pd.DataFrame, condos: bool) -> pd.DataFrame:
     """
-    Runs np.select that creates an outlier taxonomy.
+    Enhances the dataframe by adding two columns for outlier reasons:
+    - sv_char_outlier_reason: Reasons based on characteristics including price swings.
+    - sv_char_price_reason: Reasons based on high or low pricing levels.
+
     Inputs:
-        df (pd.DataFrame): dataframe with necessary columns created from previous functions.
+        df (pd.DataFrame): Dataframe with necessary columns created from previous functions.
     Outputs:
-        df (pd.DataFrame): dataframe with 'sv_outlier_type' column.
+        df (pd.DataFrame): Dataframe with 'sv_char_outlier_reason' and 'sv_char_price_reason' columns.
     """
-    if condos == True:
-        conditions = [
-            (df["sv_short_owner"] == "Short-term owner")
-            & (df["sv_pricing"].str.contains("High")),
-            (df["sv_name_match"] != "No match")
-            & (df["sv_pricing"].str.contains("High")),
-            (
-                df[["sv_buyer_category", "sv_seller_category"]]
-                .eq("legal_entity")
-                .any(axis=1)
-            )
-            & (df["sv_pricing"].str.contains("High")),
-            (df["sv_anomaly"] == "Outlier") & (df["sv_pricing"].str.contains("High")),
-            (df["sv_pricing"].str.contains("High price swing")),
-            (df["sv_pricing"].str.contains("High")),
-            (df["sv_short_owner"] == "Short-term owner")
-            & (df["sv_pricing"].str.contains("Low")),
-            (df["sv_name_match"] != "No match")
-            & (df["sv_pricing"].str.contains("Low")),
-            (
-                df[["sv_buyer_category", "sv_seller_category"]]
-                .eq("legal_entity")
-                .any(axis=1)
-            )
-            & (df["sv_pricing"].str.contains("Low")),
-            (df["sv_anomaly"] == "Outlier") & (df["sv_pricing"].str.contains("Low")),
-            (df["sv_pricing"].str.contains("Low price swing")),
-            (df["sv_pricing"].str.contains("Low")),
+    if condos:
+        # Define conditions for characteristic-based reasons
+        char_conditions = [
+            df["sv_short_owner"] == "Short-term owner",
+            df["sv_name_match"] != "No match",
+            df[["sv_buyer_category", "sv_seller_category"]]
+            .eq("legal_entity")
+            .any(axis=1),
+            df["sv_anomaly"] == "Outlier",
+            df["sv_pricing"].str.contains("High price swing"),
+            df["sv_pricing"].str.contains("Low price swing"),
         ]
 
-        labels = [
-            "Short-term owner (high)",
-            "Family sale (high)",
-            "Non-person sale (high)",
-            "Statistical Anomaly (high)",
+        # Define labels for characteristic-based reasons
+        char_labels = [
+            "Short-term owner",
+            "Family sale",
+            "Non-person sale",
+            "Statistical anomaly",
             "High price swing",
-            "High price (raw)",
-            "Short-term owner (low)",
-            "Family sale (low)",
-            "Non-person sale (low)",
-            "Statistical Anomaly (low)",
             "Low price swing",
-            "Low price (raw)",
         ]
+
+        # Define conditions for price-based reasons
+        price_conditions = [
+            df["sv_pricing"].str.contains("High"),
+            df["sv_pricing"].str.contains("Low"),
+            (df["sv_pricing"].str.contains("High")) & (df["sv_which_price"] == "(raw)"),
+            (df["sv_pricing"].str.contains("Low")) & (df["sv_which_price"] == "(raw)"),
+        ]
+
+        # Define labels for price-based reasons
+        price_labels = [
+            "High price",
+            "Low price",
+        ]
+
+        # Apply np.select to create the new columns
+        df["sv_char_outlier_reason"] = np.select(
+            char_conditions, char_labels, default="Not outlier"
+        )
+        df["sv_char_price_reason"] = np.select(
+            price_conditions, price_labels, default="Not outlier"
+        )
 
     else:
-        conditions = [
-            (df["sv_short_owner"] == "Short-term owner")
-            & (df["sv_pricing"].str.contains("High")),
-            (df["sv_name_match"] != "No match")
-            & (df["sv_pricing"].str.contains("High")),
-            (
-                df[["sv_buyer_category", "sv_seller_category"]]
-                .eq("legal_entity")
-                .any(axis=1)
-            )
-            & (df["sv_pricing"].str.contains("High")),
-            (df["sv_anomaly"] == "Outlier") & (df["sv_pricing"].str.contains("High")),
-            (df["sv_pricing"].str.contains("High price swing")),
-            (df["sv_pricing"].str.contains("High"))
-            & (df["sv_which_price"] == "(raw & sqft)"),
-            (df["sv_pricing"].str.contains("High")) & (df["sv_which_price"] == "(raw)"),
+        # Define conditions for characteristic-based reasons
+        char_conditions = [
+            df["sv_short_owner"] == "Short-term owner",
+            df["sv_name_match"] != "No match",
+            df[["sv_buyer_category", "sv_seller_category"]]
+            .eq("legal_entity")
+            .any(axis=1),
+            df["sv_anomaly"] == "Outlier",
+            df["sv_pricing"].str.contains("High price swing"),
+            df["sv_pricing"].str.contains("Low price swing"),
+        ]
+
+        # Define labels for characteristic-based reasons
+        char_labels = [
+            "Short-term owner",
+            "Family sale",
+            "Non-person sale",
+            "Statistical anomaly",
+            "High price swing",
+            "Low price swing",
+        ]
+
+        # Define conditions for price-based reasons
+        price_conditions = [
+            (df["sv_pricing"].str.contains("High") & (df["sv_which_price"] == "(raw)")),
+            (df["sv_pricing"].str.contains("Low") & (df["sv_which_price"] == "(raw)")),
             (df["sv_pricing"].str.contains("High"))
             & (df["sv_which_price"] == "(sqft)"),
-            (df["sv_short_owner"] == "Short-term owner")
-            & (df["sv_pricing"].str.contains("Low")),
-            (df["sv_name_match"] != "No match")
-            & (df["sv_pricing"].str.contains("Low")),
-            (
-                df[["sv_buyer_category", "sv_seller_category"]]
-                .eq("legal_entity")
-                .any(axis=1)
-            )
-            & (df["sv_pricing"].str.contains("Low")),
-            (df["sv_anomaly"] == "Outlier") & (df["sv_pricing"].str.contains("Low")),
-            (df["sv_pricing"].str.contains("Low price swing")),
-            (df["sv_pricing"].str.contains("Low"))
-            & (df["sv_which_price"] == "(raw & sqft)"),
-            (df["sv_pricing"].str.contains("Low")) & (df["sv_which_price"] == "(raw)"),
             (df["sv_pricing"].str.contains("Low")) & (df["sv_which_price"] == "(sqft)"),
         ]
 
-        labels = [
-            "Short-term owner (high)",
-            "Family sale (high)",
-            "Non-person sale (high)",
-            "Statistical anomaly (high)",
-            "High price swing",
-            "High price (raw & sqft)",
+        # Define labels for price-based reasons
+        price_labels = [
             "High price (raw)",
-            "High price (sqft)",
-            "Short-term owner sale (low)",
-            "Family sale (low)",
-            "Non-person sale (low)",
-            "Statistical anomaly (low)",
-            "Low price swing",
-            "Low price (raw & sqft)",
             "Low price (raw)",
+            "High price (sqft)",
             "Low price (sqft)",
         ]
 
-    df["sv_outlier_type"] = np.select(conditions, labels, default="Not outlier")
+        # Apply np.select to create the new columns
+        df["sv_outlier_reason1"] = np.select(
+            price_conditions, price_labels, default="Not outlier"
+        )
+        df["sv_outlier_reason2"] = np.select(
+            char_conditions, char_labels, default="Not outlier"
+        )
 
     return df
 
--- a/glue/flagging_script_glue/flagging.py
+++ b/glue/flagging_script_glue/flagging.py
@@ -804,16 +804,16 @@ def outlier_type(df: pd.DataFrame, condos: bool) -> pd.DataFrame:
         ]
 
         labels = [
-            "Home flip sale (high)",
+            "Short-term owner (high)",
             "Family sale (high)",
             "Non-person sale (high)",
-            "Anomaly (high)",
+            "Statistical Anomaly (high)",
             "High price swing",
             "High price (raw)",
-            "Home flip sale (low)",
+            "Short-term owner (low)",
             "Family sale (low)",
             "Non-person sale (low)",
-            "Anomaly (low)",
+            "Statistical Anomaly (low)",
             "Low price swing",
             "Low price (raw)",
         ]
@@ -856,18 +856,18 @@ def outlier_type(df: pd.DataFrame, condos: bool) -> pd.DataFrame:
         ]
 
         labels = [
-            "Home flip sale (high)",
+            "Short-term owner (high)",
             "Family sale (high)",
             "Non-person sale (high)",
-            "Anomaly (high)",
+            "Statistical anomaly (high)",
             "High price swing",
             "High price (raw & sqft)",
             "High price (raw)",
             "High price (sqft)",
-            "Home flip sale (low)",
+            "Short-term owner sale (low)",
             "Family sale (low)",
             "Non-person sale (low)",
-            "Anomaly (low)",
+            "Statistical anomaly (low)",
             "Low price swing",
             "Low price (raw & sqft)",
             "Low price (raw)",
--- a/glue/flagging_script_glue/flagging.py
+++ b/glue/flagging_script_glue/flagging.py
@@ -811,9 +811,7 @@ def outlier_type(df: pd.DataFrame, condos: bool) -> pd.DataFrame:
 
     outlier_type_dict = dict(zip(combined_labels, combined_conditions))
 
-    """
-    INDICATOR COLS 
-    """
+    # Create indicator columns for each flag type
     for label, condition in outlier_type_dict.items():
         df[label] = condition.astype(int)
 
--- a/glue/flagging_script_glue/flagging.py
+++ b/glue/flagging_script_glue/flagging.py
@@ -863,34 +863,6 @@ def outlier_type(df: pd.DataFrame, condos: bool) -> pd.DataFrame:
     return df
 
 
-"""
-def outlier_flag(df: pd.DataFrame) -> pd.DataFrame:
-
-    Creates a flag that shows whether the record is an
-    outlier (a special flag) according to our outlier taxonomy.
-    Inputs:
-        df (pd.DataFrame): dataframe to create outlier flag
-    Outputs:
-        df (pd.DataFrame): dataframe with 'is_outlier' column
-
-
-    options = ["High price", "Low price", "High price (sqft)", "Low price (sqft)"]
-    pattern = r"\b(?:" + "|".join(map(re.escape, options)) + r")\b"
-
-    df["sv_is_outlier"] = np.select(
-        [
-            df["sv_outlier_reason1"].str.contains(
-                pattern, case=False, na=False, regex=True
-            )
-        ],
-        [1],
-        default=0,
-    )
-
-    return df
-"""
-
-
 # STRING CLEANUP
 
 """
--- a/glue/flagging_script_glue/flagging.py
+++ b/glue/flagging_script_glue/flagging.py
@@ -598,11 +598,11 @@ def get_sale_counts(dups: pd.DataFrame) -> pd.DataFrame:
     v_counts = (
         dups.pin.value_counts()
         .reset_index()
-        .rename(columns={"index": "pin", "pin": "sv_sale_dup_counts"})
+        .rename(columns={"count": "sv_sale_dup_counts"})
     )
 
     # Explicitly specify the merging columns
-    dups = pd.merge(dups, v_counts, on="pin")
+    dups = pd.merge(dups, v_counts)
 
     return dups
 
@@ -803,8 +803,6 @@ def outlier_type(df: pd.DataFrame, condos: bool) -> pd.DataFrame:
         price_conditions = [
             df["sv_pricing"].str.contains("High"),
             df["sv_pricing"].str.contains("Low"),
-            (df["sv_pricing"].str.contains("High")) & (df["sv_which_price"] == "(raw)"),
-            (df["sv_pricing"].str.contains("Low")) & (df["sv_which_price"] == "(raw)"),
         ]
 
         # Define labels for price-based reasons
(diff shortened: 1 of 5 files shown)


manual_flagging/flagging.py
```diff
@@ -315,18 +315,21 @@ def create_bins_and_labels(input_list):
...
     )
+
+    df_copy = flg.group_size_adjustment(
+        df=df_copy,
+        stat_groups=df_info["columns"],
+        min_threshold=inputs["min_groups_threshold"],
+        condos=df_info["condos_boolean"],
+    )
+
     """
```
wagnerlmichael, line 333: This is the second piece of problematic code with regards to the false positives. The group size adjustment seems to have been happening before the ptax outliers were assigned. The actual assignment of `sv_is_outlier` to `1` happened in `finish_flags()`, which also ran after the `group_size adjustment` The fix in this PR puts all of the ptax handling directly in the `ptax_adjustment()` function and swaps the order of the `ptax_adjustment` and `group_size_adjustment()`.

glue/flagging_script_glue/flagging.py
```diff
@@ -767,114 +767,125 @@ def z_normalize_groupby(s: pd.Series):
...
+    def find_next_unused_char(row, df):
+        used_labels = {row["sv_outlier_reason1"], row["sv_outlier_reason2"]}
+        for cond, label in zip(char_conditions, char_labels):
+            # Evaluate the condition directly for the current row
+            if label not in used_labels and cond[row.name]:
+                return label
+        return "Not outlier"
+
+    # Now we apply this function row-wise correctly by also passing the DataFrame reference:
+    df["sv_outlier_reason3"] = df.apply(find_next_unused_char, df=df, axis=1)
```
wagnerlmichael, line 888: Here we look to select a price outlier first (price is picked first by order of the `combined_conditions` and `combined_labels`) and if there are no price outliers, we assign a characteristic reason to populate `sv_outlier_reason1`. If this is the case, the observation won't be labeled as `1` for `sv_is_outlier` since we need a price outlier match in order to classify it as an outlier. However, the characteristic reason will remain for the `sv_outlier_reason1` column. 

Then, we take the next label that hasn't been already placed in `sv_outlier_reason1` and place it in `sv_outlier_reason2`. This could be another price label (raw and then sqft) or a characteristic option.  

For `sv_outlier_reason3` we look to see if there is another characteristic condition that has been satisfied.

glue/flagging_script_glue/flagging.py
```diff
@@ -767,114 +767,125 @@ def z_normalize_groupby(s: pd.Series):
...
+            .any(axis=1),
+            df["sv_anomaly"] == "Outlier",
+            df["sv_pricing"].str.contains("High price swing"),
+            df["sv_pricing"].str.contains("Low price swing"),
+        ]
+
+        # Define labels for characteristic-based reasons
+        char_labels = [
+            "Short-term owner",
+            "Family sale",
```
wagnerlmichael, line 838: We disentangle price and characteristic outlier reasons/labels to fit with the new specification.

glue/flagging_script_glue/flagging.py
```diff
@@ -598,11 +598,11 @@ def get_sale_counts(dups: pd.DataFrame) -> pd.DataFrame:
     v_counts = (
         dups.pin.value_counts()
         .reset_index()
-        .rename(columns={"index": "pin", "pin": "sv_sale_dup_counts"})
+        .rename(columns={"count": "sv_sale_dup_counts"})
     )
 
     # Explicitly specify the merging columns
-    dups = pd.merge(dups, v_counts, on="pin")
+    dups = pd.merge(dups, v_counts)
```
wagnerlmichael, line 605: This change was needed to work with the current pandas version, I believe. It would through an error with the previous code.

glue/flagging_script_glue/flagging.py
```diff
@@ -889,8 +900,19 @@ def outlier_flag(df: pd.DataFrame) -> pd.DataFrame:
...
+
     df["sv_is_outlier"] = np.select(
-        [(df["sv_outlier_type"] == "Not outlier")], [0], default=1
+        [
+            df["sv_outlier_reason1"].str.contains(
+                pattern, case=False, na=False, regex=True
+            )
+        ],
+        [1],
+        default=0,
```
wagnerlmichael, line 915: This code is where we actually decide whether to label an observation as an outlier based on the new `sv_outlier_reason$n` columns.  

For an observation to be flagged as an outlier it needs at least to be a price outlier. In `outlier_type()` the logic puts the price outliers by default in the first column `sv_outlier_reason1`. If there is a price flag - it will be found in this column. So to check if it is an outlier we match on that column with any of the 4 potential price outlier options.

glue/sales_val_flagging.py
```diff
@@ -131,6 +131,54 @@ def ptax_adjustment(df, groups, ptax_sd, condos: bool):
...
+            "PTAX-203", na=False
+        ),
+        # Heuristics flagging binary column
+        sv_is_heuristic_outlier=lambda df: (
+            ~df["sv_outlier_reason1"].str.contains("PTAX-203", na=False)
+        )
+        & (df["sv_is_outlier"] == 1),
+        sv_is_outlier=lambda df: df["sv_is_outlier"].astype(int),
+    )
+
```
wagnerlmichael, line 181: Two things are happening in this change.

1. We move some ptax processing code from `finish_flags()` into `ptax_adjustment()`. This improves organization - instead of ptax processing split across multiple functions, we now see most of the ptax processing under one function called `ptax_adjustment()`
2. We handle the ptax processing with the new design doc specifications. If the observation is a ptax outlier, then we assign it for the `sv_outlier_reason1`, shift the existing `sv_outlier_reason1` to `sv_outlier_reason2`, and shift the existing `sv_outlier_reason2` to `sv_outlier_reason3`.


First we take the old column and overwrite the `sv_outlier_reason1` if it is a ptax outlier.
```python
new_sv_outlier_reason1 = np.select(
        [
            (df["ptax_flag_w_deviation"]) & (df["ptax_direction"] == "High"),
            (df["ptax_flag_w_deviation"]) & (df["ptax_direction"] == "Low"),
        ],
        ["PTAX-203 Exclusion (High)", "PTAX-203 Exclusion (Low)"],
        default=df["sv_outlier_reason1"],
    )
```

Then if the following conditions are fulfilled (code and written)
```python
(new_sv_outlier_reason1 != df["sv_outlier_reason1"])
            & (df["sv_outlier_reason1"] != "Not outlier")
```
- the `new_sv_outlier_reason1` and the original `sv_outlier_reason1` aren't equal (due to ptax overwrite)
- the original `sv_outlier_reason1` isn't `Not outlier`

we shift the `sv_outlier_reason1` to `sv_outlier_reason2`. We do the same from `sv_outlier_reason2` to `sv_outlier_reason3`.

glue/flagging_script_glue/flagging.py
```diff
@@ -767,114 +767,156 @@ def z_normalize_groupby(s: pd.Series):
...
+            # Move the current sv_outlier_reason2 to sv_outlier_reason3
+            row["sv_outlier_reason3"] = row["sv_outlier_reason2"]
+
+            # Update sv_outlier_reason2 to the new value '(sqft)'
+            row["sv_outlier_reason2"] = f"{price_type} (sqft)"
+
+        return row
+
+    # Apply the transformation
+    df = df.apply(transform_row, axis=1)
```
wagnerlmichael, line 919: This helper function was necessary because the mansueto model outputs "raw and sqft" price outlier information together, and we want to separate them based on the design doc specification. I looked into decoupling this but I think it would have been a massive diff and a lot of work. I think this helper function is the most simple solution.

glue/flagging_script_glue/flagging.py
```diff
@@ -889,8 +931,17 @@ def outlier_flag(df: pd.DataFrame) -> pd.DataFrame:
...
+
     df["sv_is_outlier"] = np.select(
-        [(df["sv_outlier_type"] == "Not outlier")], [0], default=1
+        [
+            df["sv_outlier_reason1"].str.contains(
+                pattern, case=False, na=False, regex=True
+            )
+        ],
+        [1],
+        default=0,
```
wagnerlmichael, line 944: Regex matching here to assign outlier status. This works because we need a price outlier for the observation to be classified as an outleir.

glue/sales_val_flagging.py
```diff
@@ -131,6 +131,86 @@ def ptax_adjustment(df, groups, ptax_sd, condos: bool):
...
+        sv_is_ptax_outlier=lambda df: df["sv_outlier_reason1"].str.contains(
+            "PTAX-203", na=False
+        ),
+        # Heuristics flagging binary column
+        sv_is_heuristic_outlier=lambda df: (
+            ~df["sv_outlier_reason1"].str.contains("PTAX-203", na=False)
+        )
+        & (df["sv_is_outlier"] == 1),
+        sv_is_outlier=lambda df: df["sv_is_outlier"].astype(int),
+    )
```
wagnerlmichael, line 212: Two things are happening in this change.

1. We move some ptax processing code from `finish_flags()` into `ptax_adjustment()`. This improves organization - instead of ptax processing split across multiple functions, we now see most of the ptax processing under one function called `ptax_adjustment()`
2. We handle the ptax processing with the new design doc specifications. If the observation is a ptax outlier, then we assign it for the `sv_outlier_reason1`, shift the existing `sv_outlier_reason1` to `sv_outlier_reason2`, and shift the existing `sv_outlier_reason2` to `sv_outlier_reason3`.


First we take the old column and overwrite the `sv_outlier_reason1` if it is a ptax outlier.
```python
new_sv_outlier_reason1 = np.select(
        [
            (df["ptax_flag_w_deviation"]) & (df["ptax_direction"] == "High"),
            (df["ptax_flag_w_deviation"]) & (df["ptax_direction"] == "Low"),
        ],
        ["PTAX-203 Exclusion (High)", "PTAX-203 Exclusion (Low)"],
        default=df["sv_outlier_reason1"],
    )
```

Then if the following conditions are fulfilled (code and written)
```python
(new_sv_outlier_reason1 != df["sv_outlier_reason1"])
            & (df["sv_outlier_reason1"] != "Not outlier")
```
- the `new_sv_outlier_reason1` and the original `sv_outlier_reason1` aren't equal (due to ptax overwrite)
- the original `sv_outlier_reason1` isn't `Not outlier`

we shift the `sv_outlier_reason1` to `sv_outlier_reason2`. We do the same from `sv_outlier_reason2` to `sv_outlier_reason3`.

Finally, we infill the original `sv_outlier_reason1` with `new_sv_outlier_reason1`
```python
.assign(
        sv_outlier_reason1=new_sv_outlier_reason1 
    )
```

manual_flagging/requirements.txt
```diff
@@ -1,4 +1,4 @@
-awswrangler==2.20.1
+awswrangler==3.7.3
```
wagnerlmichael, line 1: Bumped this because I was getting some sort of parquet error. It fixed the issue.

glue/flagging_script_glue/flagging.py
```diff
@@ -767,114 +767,156 @@ def z_normalize_groupby(s: pd.Series):
...
+            df["sv_short_owner"] == "Short-term owner",
+            df["sv_name_match"] != "No match",
+            df[["sv_buyer_category", "sv_seller_category"]]
+            .eq("legal_entity")
+            .any(axis=1),
+            df["sv_anomaly"] == "Outlier",
+            df["sv_pricing"].str.contains("High price swing"),
+            df["sv_pricing"].str.contains("Low price swing"),
+        ]
+
```
wagnerlmichael, line 835: Disentangling price and char outlier reasons to fit with new specification.

glue/sales_val_flagging.py
```diff
@@ -131,10 +135,62 @@ def ptax_adjustment(df, groups, ptax_sd, condos: bool):
...
+
```
wagnerlmichael, line 189: Two things are happening in this change.

1. We move some ptax processing code from `finish_flags()` into `ptax_adjustment()`. This improves organization - instead of ptax processing split across multiple functions, we now see most of the ptax processing under one function called `ptax_adjustment()`
2. We handle the ptax processing with the new design doc specifications. If the observation is a ptax outlier, then we assign it for the `sv_outlier_reason1`, shift the existing `sv_outlier_reason1` to `sv_outlier_reason2`, and shift the existing `sv_outlier_reason2` to `sv_outlier_reason3`.


First we take the old column and overwrite the `sv_outlier_reason1` if it is a ptax outlier. This overwrite also happened in the previous version of the pipeline.
```python
new_sv_outlier_reason1 = np.select(
        [
            (df["ptax_flag_w_deviation"]) & (df["ptax_direction"] == "High"),
            (df["ptax_flag_w_deviation"]) & (df["ptax_direction"] == "Low"),
        ],
        ["PTAX-203 Exclusion (High)", "PTAX-203 Exclusion (Low)"],
        default=df["sv_outlier_reason1"],
    )
```

Then if the following conditions are fulfilled (code and written)
```python
(new_sv_outlier_reason1 != df["sv_outlier_reason1"])
            & (df["sv_outlier_reason1"] != "Not outlier")
```
- the `new_sv_outlier_reason1` and the original `sv_outlier_reason1` aren't equal (due to ptax overwrite)
- the original `sv_outlier_reason1` isn't `Not outlier`

we shift the `sv_outlier_reason1` to `sv_outlier_reason2`. We do the same from `sv_outlier_reason2` to `sv_outlier_reason3`.

Finally, we infill the original `sv_outlier_reason1` with `new_sv_outlier_reason1`
```python
.assign(
        sv_outlier_reason1=new_sv_outlier_reason1 
    )
```

glue/flagging_script_glue/flagging.py
```diff
@@ -767,114 +767,156 @@ def z_normalize_groupby(s: pd.Series):
...
+    )
+
+    # Adjust for reason2 based on remaining price conditions and char conditions not used in reason1
+    remaining_conditions = [
+        (cond & (df["sv_outlier_reason1"] != label))
+        for cond, label in zip(combined_conditions, combined_labels)
+    ]
+    df["sv_outlier_reason2"] = np.select(
+        remaining_conditions, combined_labels, default="Not outlier"
+    )
```
jeancochrane, line 884: **[Thought, non-blocking]** My intuition says that there's got to be a way to simplify these filter operations such that we don't need completely different logic in order to set reasons 1, 2, and 3. In my ideal imagined world, we would just have one list of conditions in descending order based on priority, and then we would select all the conditions that match each row and keep the top three. Does that seem feasible, or is there a strong reason to prefer we do it this way? This is another thing we could discuss in person!

jeancochrane (commented):
This is a great start! Sounds like we missed the `Price swing / home flip` characteristic, so I'll hold off on approving until that's in and I can take a look at the whole thing.

glue/flagging_script_glue/flagging.py
```diff
@@ -767,114 +767,156 @@ def z_normalize_groupby(s: pd.Series):
...
+            df["sv_name_match"] != "No match",
+            df[["sv_buyer_category", "sv_seller_category"]]
+            .eq("legal_entity")
+            .any(axis=1),
+            df["sv_anomaly"] == "Outlier",
+            df["sv_pricing"].str.contains("High price swing"),
+            df["sv_anomaly"] == "Outlier",
+            df["sv_pricing"].str.contains("High price swing")
+            | df["sv_pricing"].str.contains("Low price swing"),
+        ]
```
jeancochrane, line 835: **[Question, non-blocking]** It seems like res and condos share the same set of `char_conditions`, is that right? If so, perhaps we could define one `char_conditions` list to be shared by both workflows, and then have `price_conditions` be distinct depending on the value of `condos`?

glue/flagging_script_glue/flagging.py
```diff
@@ -270,26 +266,6 @@ def pricing_info(
...
-        (df["sv_short_owner"] == "Short-term owner"),
-        (df["sv_transaction_type"] == "legal_entity-legal_entity"),
-    ]
-    labels = ["Family sale", "Home flip sale", "Non-person sale"]
-
-    df["sv_special_flags"] = np.select(cond, labels, default="Not special")
-
-    return df
-
-
```
wagnerlmichael, line 292: As far as I could tell, this function was serving no purpose in the pipeline, and I couldn't find any references to the `sv_special_flags` column in the repo.

glue/flagging_script_glue/flagging.py
```diff
@@ -767,131 +742,97 @@ def z_normalize_groupby(s: pd.Series):
...
+            ["sv_ind_price_low_price_sqft", "sv_ind_price_low_price"],
+        ] = 1
+
+        df.drop(
+            columns=[
+                "sv_ind_price_high_price_raw_and_sqft",
+                "sv_ind_price_low_price_raw_and_sqft",
+            ],
+            inplace=True,
+        )
```
wagnerlmichael, line 835: I didn't find a simple way to disentangle this from the prior functions so I had to place this logic here to separate raw and sqft classification.

glue/sales_val_flagging.py
```diff
@@ -102,45 +102,42 @@ def ptax_adjustment(df, groups, ptax_sd, condos: bool):
     group_string = "_".join(groups)
 
     if not condos:
-        df["ptax_flag_w_deviation"] = df["ptax_flag_original"] & (
+        df["sv_ind_ptax_flag_w_deviation"] = df["ptax_flag_original"] & (
```
wagnerlmichael, line 105: Change this to be in line with new indicator column naming convention

glue/sales_val_flagging.py
```diff
@@ -152,6 +149,66 @@ def group_size_adjustment(df, stat_groups: list, min_threshold, condos: bool):
...
+        "High price per square foot",
+        "Low price per square foot",
+    }
+    df["sv_is_outlier"] = np.where(df["sv_outlier_reason1"].isin(values_to_check), 1, 0)
+
+    # - - -
+    # Handle group threshold adjustment
+    # TODO: Explain choice of sv_outlier_reason$n, do we set all reasons to 0 as well?
+    # - - -
+
```
wagnerlmichael, line 211: This is an open question I want some advice on. Currently when we are manually setting the outlier status from outiler to non-outlier based on the group threshold, we only change `sv_is_outlier` from 1 to 0. We leave all of the `sv_outlier_reason$n` columns as they are. So they could have a 'High price' and a 'Family sale' status and still be 0 for `sv_is_outlier`. I was leaning towards leaving this as it is still potentially useful information, but I could see it as potentially confusing.

glue/sales_val_flagging.py
```diff
@@ -193,7 +235,18 @@ def group_size_adjustment(df, stat_groups: list, min_threshold, condos: bool):
...
+        sv_is_ptax_outlier=lambda df: df["sv_ind_ptax_flag_w_deviation"],
+        na=False,
+        sv_is_heuristic_outlier=lambda df: (
+            ~df["sv_outlier_reason1"].str.contains("PTAX-203", na=False)
+        )
+        & (df["sv_is_outlier"] == 1),
+        sv_is_outlier=lambda df: df["sv_is_outlier"].astype(int),
+    )
+
+    return df
```
wagnerlmichael, line 249: This was moved here from `finish_flags`

glue/sales_val_flagging.py
```diff
@@ -226,45 +279,17 @@ def finish_flags(df, start_date, manual_update, sales_to_write_filter):
...
-            # PTAX-203 binary
-            sv_is_ptax_outlier=lambda df: df["sv_outlier_type"] == "PTAX-203 flag",
-            # Heuristics flagging binary column
-            sv_is_heuristic_outlier=lambda df: (
-                df["sv_outlier_type"] != "PTAX-203 flag"
-            )
-            & (df["sv_is_outlier"] == 1),
-        )
-    )
-
```
wagnerlmichael, line 258: This was broken up into `classify_outliers()` and `ptax_adjustment()`

glue/sales_val_flagging.py
```diff
@@ -152,6 +149,63 @@ def group_size_adjustment(df, stat_groups: list, min_threshold, condos: bool):
...
+                reason = col  # Use column name as reason
+                # Check if reason is not already recorded and there's a 'Null' slot to fill
+                if reason not in reasons and "Null" in reasons:
+                    next_index = reasons.index("Null") + 1  # Find the next 'Null' index
+                    row[f"sv_outlier_reason{next_index}"] = reason
+                    reasons[next_index - 1] = reason  # Update reasons list
+
+        return row
+
+    df = df.apply(fill_outlier_reasons, axis=1)
```
wagnerlmichael, line 176: Forward fill the reason columns with ptax first, then price, then char reasons.

glue/sales_val_flagging.py
```diff
@@ -152,6 +149,63 @@ def group_size_adjustment(df, stat_groups: list, min_threshold, condos: bool):
...
+        "sv_ind_price_high_price": "High price",
+        "sv_ind_price_low_price": "Low price",
+        "sv_ind_price_high_price_sqft": "High price per square foot",
+        "sv_ind_price_low_price_sqft": "Low price per square foot",
+        "sv_ind_char_short_term_owner": "Short-term owner",
+        "sv_ind_char_family_sale": "Family Sale",
+        "sv_ind_char_non_person_sale": "Non-person sale",
+        "sv_ind_char_statistical_anomaly": "Statistical Anomaly",
+        "sv_ind_char_price_swing_homeflip": "Price swing / Home flip",
+    }
```
wagnerlmichael, line 189: Translate indicator cols to more readable / design doc specification.

glue/sales_val_flagging.py
```diff
@@ -152,6 +149,63 @@ def group_size_adjustment(df, stat_groups: list, min_threshold, condos: bool):
...
+
+    # Assign outlier status
+    values_to_check = {
+        "PTAX-203 Exclusion",
+        "High price",
+        "Low price",
+        "High price per square foot",
+        "Low price per square foot",
+    }
+    df["sv_is_outlier"] = np.where(df["sv_outlier_reason1"].isin(values_to_check), 1, 0)
```
wagnerlmichael, line 202: Assign the actual outlier flag here.

glue/flagging_script_glue/flagging.py
```diff
@@ -767,131 +742,97 @@ def z_normalize_groupby(s: pd.Series):
...
+            "sv_ind_price_low_price_sqft",
         ]
 
-    df["sv_outlier_type"] = np.select(conditions, labels, default="Not outlier")
-
-    return df
+    combined_conditions = price_conditions + char_conditions
+    combined_labels = price_labels + char_labels
 
+    outlier_type_dict = dict(zip(combined_labels, combined_conditions))
```
jeancochrane, line 812: **[Nitpick, non-blocking]** With my previous comment (https://github.com/ccao-data/model-sales-val/pull/128#discussion_r1627990344), I was imagining that instead of zipping these lists and casting them to a dict, we might just combine them into a single data structure at the point of definition; that way we wouldn't need to bother with this extra step, and it would be easier for readers to quickly understand which conditions map to which labels without having to compare two lists by index. That being said, I don't think it's actually a big deal, and I'm fine going forward with whatever approach you prefer -- I just want to make sure the intention of my comment was clear! If we decide to move forward with two lists that we zip together, as in this implementation, I don't think the extra `dict()` call is necessary since the zipped lists can be iterated on their own.

glue/sales_val_flagging.py
```diff
@@ -102,45 +102,42 @@ def ptax_adjustment(df, groups, ptax_sd, condos: bool):
...
-        # Determine the ptax direction
-        conditions = [
-            (df[f"sv_price_deviation_{group_string}"] >= ptax_sd[1]),
-            (df[f"sv_price_deviation_{group_string}"] <= -ptax_sd[0]),
-        ]
+    df["sv_ind_ptax_flag_w_deviation"] = df["sv_ind_ptax_flag_w_deviation"].astype(int)
 
-    directions = ["High", "Low"]
-    df["ptax_direction"] = np.select(conditions, directions, default=np.nan)
+    # Adjust order to make classify_outliers function smoother
```
jeancochrane, line 120: **[Nitpick, non-blocking]** A small recommended change to this comment, since I misinterpreted it on first read:

```suggestion
    # Adjust order of indicator columns to make classify_outliers function smoother
```

But also, per my comment below, perhaps there's a way to better define priority such that the order of these columns is not important.

glue/sales_val_flagging.py
```diff
@@ -152,6 +149,63 @@ def group_size_adjustment(df, stat_groups: list, min_threshold, condos: bool):
         df: dataframe with newly manually adjusted outlier values.
 
     """
+
+    # Assign sv_outlier_reasons
+    df["sv_outlier_reason1"] = "Null"
+    df["sv_outlier_reason2"] = "Null"
+    df["sv_outlier_reason3"] = "Null"
```
jeancochrane, line 156: **[Question, non-blocking]** Why set these to the string `"Null"` instead of `np.nan`?

glue/sales_val_flagging.py
jeancochrane: **[Suggestion, non-blocking]** If we're already creating a crosswalk data structure for all of the reason codes, I think we should use it instead of `cols_to_iterate` as the canonical list of reason codes for forward-filling. That way we can be explicit about the intended order of priority for the columns, instead of relying on the implicit order of the columns in the dataframe to establish priority (dicts are ordered starting in Python 3.7, which is the same version used by [Glue 2.0](https://docs.aws.amazon.com/glue/latest/dg/glue-version-support-policy.html)).

glue/sales_val_flagging.py
jeancochrane: Hmm, this is a good question! I think we should default to preserving all of the fields, but maybe what we need to make things clearer is an additional reason code like "Group size too small to flag" that we can add in cases like these. I'll raise this question with Dan and we can settle on a path forward.

glue/sales_val_flagging.py
```diff
@@ -193,7 +232,18 @@ def group_size_adjustment(df, stat_groups: list, min_threshold, condos: bool):
         lambda row: "_".join([str(row[col]) for col in stat_groups]), axis=1
     )
 
-    return df_flagged_updated
+    df = df_flagged_updated.assign(
+        # PTAX-203 binary
+        sv_is_ptax_outlier=lambda df: df["sv_ind_ptax_flag_w_deviation"],
+        na=False,
```
jeancochrane, line 238: **[Question, non-blocking]** Am I right that this adds a `na` column to `df`? What's its purpose?

glue/sales_val_flagging.py
```diff
@@ -193,7 +232,18 @@ def group_size_adjustment(df, stat_groups: list, min_threshold, condos: bool):
...
+        sv_is_heuristic_outlier=lambda df: (
+            ~df["sv_outlier_reason1"].str.contains("PTAX-203", na=False)
+        )
```
jeancochrane, line 241: **[Nitpick, non-blocking]** This expression relies on the assumption that PTAX-203 flags always get highest priority in the list of reasons. Perhaps a more robust solution would just be to check if `df["sv_in_ptax_flag_w_deviation"]` is not 1?

glue/flagging_script_glue/flagging.py
jeancochrane: **[Question, non-blocking]** Hmm, I'm confused why this needs a separate step. Can't we just skip e.g. the `sv_ind_price_high_price_raw_and_sqft` condition and use two separate conditions to make this extra step unnecessary?

1.  `df["sv_ind_price_high_price"] = df["sv_pricing"].str.contains("High")`
2.  `df["sv_ind_price_high_price_sqft"] = df["sv_pricing"].str.contains("High") & df["sv_pricing"].str.contains("sqft")`

jeancochrane (approved):
Looking great! This is a big improvement over the last iteration. A few ideas for you on further cleanup below, but I think this is in a state where I would be fine with merging it as-is if need be.

Seeing that there are two places where we're defining outlier type metadata (i.e. the list of conditions in `outlier_type` and the crosswalk in `classify_outliers`) is making me wonder whether it would be worthwhile to consolidate all of the different outlier metadata into one data structure that is defined at the top of the module and shared between all of the functions. This data structure could then own all of the outlier type configuration, like conditions, labels, priority order, and whether or not the condition should indicate that `sv_is_outlier` be flipped to 1. Something like this:

```python
OUTLIER_TYPES = [
    {
        "colname": "sv_ind_ptax_flag_w_deviation",
        "label": "PTAX-203 Exclusion",
        "condition": {
            "res": lambda df: df["ptax_flag_original"] & (
                (df[f"sv_price_deviation_{group_string}"] >= ptax_sd[1])
                | (df[f"sv_price_deviation_{group_string}"] <= -ptax_sd[0])
                | (df[f"sv_price_per_sqft_deviation_{group_string}"] >= ptax_sd[1])
                | (df[f"sv_price_per_sqft_deviation_{group_string}"] <= -ptax_sd[0])
            ),
           "condo": lambda df: df["ptax_flag_original"] & (
                (df[f"sv_price_deviation_{group_string}"] >= ptax_sd[1])
                | (df[f"sv_price_deviation_{group_string}"] <= -ptax_sd[0])
            )
        },
        "determines_outlier": True,
    },
    {
        "colname": "sv_ind_price_high_price",
        "label": "High price",
        "condition": {
            "res": lambda df: df["sv_pricing"].str.contains("High"),
            "condo": lambda df: df["sv_pricing"].str.contains("High"),
        },
        "determines_outlier": True,
    },
    ...
]
```

Note that the two examples above both have `determines_outlier=True`, but for e.g. `Family sale` we would set `determines_outlier=False` such that the presence of that flag doesn't automatically set `sv_is_outler=1`. We could also imagine altering the data structure such that `condition.res` and `condition.condo` are only set if the outlier type requires different conditions for the two property types, whereas`condition` is set to a single callable if a condition is shared between the two property types (as in `High price`), but that's more of a personal preference.

This kind of unified data structure is not strictly necessary, but I think it could help us clean up the code even further! Curious what you think, and happy to jump on a call to explain in more detail.

glue/sales_val_flagging.py
wagnerlmichael: Yes, good catch! Not sure why that is in there. I must have left it from older code.

wagnerlmichael:
https://github.com/ccao-data/model-sales-val/pull/128#discussion_r1633546545 This was a great suggestion. I implemented a fix here: https://github.com/ccao-data/model-sales-val/pull/128/commits/38b99c7b1f3b252086e52623d33cec49de6acde0. You were right, but instead of `sv_pricing` we needed to check `sv_which_price` as that is where the sqft info is

glue/sales_val_flagging.py
wagnerlmichael: For some reason I thought it would be more intuitive for analysts. I switched it to nan as that probably plays better with the import / our data lake.

wagnerlmichael:
Placing a template for an overarching outlier type data structure here. We can come back to this whenever we do a large re-factor of the pipeline.

```python 





OUTLIER_TYPES = [
    {
        "colname": "sv_ind_ptax_flag_w_deviation",
        "label": "PTAX-203 Exclusion",
        "condition": {
            # TODO: include cast to int
            "res": lambda df, context: df["ptax_flag_original"] & (
                (df[f"sv_price_deviation_{context['group_string']}"] >= ptax_sd[1])
                | (df[f"sv_price_deviation_{group_string}"] <= -ptax_sd[0])
                | (df[f"sv_price_per_sqft_deviation_{group_string}"] >= ptax_sd[1])
                | (df[f"sv_price_per_sqft_deviation_{group_string}"] <= -ptax_sd[0])
            ),
           "condo": lambda df: df["ptax_flag_original"] & (
                (df[f"sv_price_deviation_{group_string}"] >= ptax_sd[1])
                | (df[f"sv_price_deviation_{group_string}"] <= -ptax_sd[0])
            )
        },
        "determines_outlier": True,
    },
    {
        "colname": "sv_ind_price_high_price",
        "label": "High price",
        "condition": {
            "res": lambda df: (
                df["sv_pricing"].str.contains("High")
                & (df["sv_which_price"].str.contains("raw"))
            ),
            "condo": lambda df: df["sv_pricing"].str.contains("High"),
        },
        "determines_outlier": True,
    },
    {
        "colname": "sv_ind_price_low_price",
        "label": "Low price",
        "condition": {
            "res": (
                lambda df: df["sv_pricing"].str.contains("Low")
                & (df["sv_which_price"].str.contains("raw"))
            ),
            "condo":lambda df: df["sv_pricing"].str.contains("Low"),
        },
        "determines_outlier": True,
    },
    {
        "colname": "sv_ind_price_high_price_sqft",
        "label": "High price per square foot",
        "condition": {
            "res": lambda df: (
                df["sv_pricing"].str.contains("High"))
                & (df["sv_which_price"].str.contains("sqft")
            ),
        },
        "determines_outlier": True,
    },
    {
        "colname": "sv_ind_price_low_price_sqft",
        "label": "Low price per square foot",
        "condition": {
            "res": lambda df: (
                df["sv_pricing"].str.contains("Low"))
            & (df["sv_which_price"].str.contains("sqft")
            )
        },
        "determines_outlier": True,
    },
    {
        "colname": "sv_ind_char_short_term_owner",
        "label": "Short-term owner",
        "condition": lambda df: df["sv_short_owner"] == "Short-term owner",
        "determines_outlier": False,
    },
    {
        "colname": "sv_ind_char_family_sale",
        "label": "Family sale",
        "condition": lambda df: df["sv_name_match"] != "No match",
        "determines_outlier": False,
    },
    {
        "colname": "sv_ind_char_non_person_sale",
        "label": "Non-person sale",
        "condition": lambda df: df[["sv_buyer_category", "sv_seller_category"]].eq("legal_entity").any(axis=1),
        "determines_outlier": False,
    },
    {
        "colname": "sv_ind_char_statistical_anomaly",
        "label": "Statistical Anomaly",
        "condition": lambda df: df["sv_anomaly"] == "Outlier",
        "determines_outlier": False,
    },
    {
        "colname": "sv_ind_char_price_swing_homeflip",
        "label": "Price swing / Home flip",
        "condition":  lambda df: df["sv_pricing"].str.contains("High price swing")
        | df["sv_pricing"].str.contains("Low price swing"),
        "determines_outlier": False,
    },
]

```

glue/sales_val_flagging.py
wagnerlmichael: I now utilize this `outlier_type_crosswalk` for the operations in this function, and delete the previous order handling in `ptax_adjustment`. Less redundancy and a defined order rather than relying on column order.

glue/flagging_script_glue/flagging.py
wagnerlmichael: I removed the dict creation and just zipped them for now. This function is at least a lot more organized and concise than the previous iteration.

glue/sales_val_flagging.py
```diff
@@ -152,6 +144,63 @@ def group_size_adjustment(df, stat_groups: list, min_threshold, condos: bool):
...
+                reason = col  # Use column name as reason
+                # Check if reason is not already recorded and there's a np.nan slot to fill
+                if reason not in reasons and pd.isna(reasons).any():
+                    next_index = pd.isna(
+                        reasons
+                    ).argmax()  # Find the index of the first True (NaN)
+                    row[f"sv_outlier_reason{next_index + 1}"] = reason
+                    reasons[next_index] = reason  # Update reasons array
+
+        return row
```
jeancochrane, line 182: **[Suggestion, non-blocking]** I think we could simplify this a bit based on the following assumptions:

1. All `sv_outlier_reason` columns will be NA at the start of this block, because we set them that way on lines 149-151
2. All keys in `outlier_type_crosswalk` are unique

If either of these two assumptions are false, let me know and disregard this suggestion; otherwise, we could try this:

```suggestion
    def fill_outlier_reasons(row):
        reason_idx = 1
        for col in outlier_type_crosswalk:
            if row[col]:  # If the condition is True
                reason = col  # Use column name as reason
                row[f"sv_outlier_reason{reason_idx}"] = reason
                if reason_idx >= 3:
                    break
                reason_idx += 1

        return row
```

glue/sales_val_flagging.py
```diff
@@ -152,6 +144,63 @@ def group_size_adjustment(df, stat_groups: list, min_threshold, condos: bool):
...
+
+    # Assign outlier status
+    values_to_check = {
+        "PTAX-203 Exclusion",
+        "High price",
+        "Low price",
+        "High price per square foot",
+        "Low price per square foot",
+    }
+    df["sv_is_outlier"] = np.where(df["sv_outlier_reason1"].isin(values_to_check), 1, 0)
```
jeancochrane, line 197: **[Suggestion, non-blocking]** It might be nice to check if _any_ reason matches `values_to_check`, in case we decide to change the priority ordering later, but I don't think it's particularly high priority if you're not concerned about that!

```suggestion
    df["sv_is_outlier"] = np.where(
        df[
            ["sv_outlier_reason{idx}" for idx in range(1, 4)]
        ].isin(values_to_check).any(axis=1), 
        1, 
        0
    )
```

jeancochrane (approved):
Looking excellent! I have a few tiny suggestions for further code simplification, but they're all optional. We just need to get a path forward on [the non-outlier-with-flags edge case](https://github.com/ccao-data/model-sales-val/pull/128/files#r1631453067) and then we should be good to go.

glue/sales_val_flagging.py
```diff
@@ -152,30 +144,98 @@ def group_size_adjustment(df, stat_groups: list, min_threshold, condos: bool):
...
+                and row[reason]
+                and not (row["_merge"] == "both" and reason in group_thresh_price_fix)
+            ):
+                row[f"sv_outlier_reason{reason_idx}"] = outlier_type_crosswalk[reason]
+                if reason_idx >= 3:
+                    break
+                reason_idx += 1
+        return row
+
+    df = df.apply(fill_outlier_reasons, axis=1)
```
wagnerlmichael, line 199: This is how we I handled the implementation of not counting the price outliers but retaining the char outliers given a group threshold violation.

The `row["_merge"] == "both"` check checks for the group thresh violation, and the `reason in group_thresh_price_fix` checks to see if we are looking at a price indicator column. So it basically says:

if not ( below threshold and price outlier) skip this iteration in the loop, which means the condition won't be filled in. 

I thought this approach would be better than manually re-arranging at the end of the function.

glue/sales_val_flagging.py
```diff
@@ -152,30 +156,119 @@ def group_size_adjustment(df, stat_groups: list, min_threshold, condos: bool):
...
+
+    outlier_type_crosswalk = {
+        "sv_ind_price_high_price": "High price",
+        "sv_ind_ptax_flag_w_high_price": "High price",
+        "sv_ind_price_low_price": "Low price",
+        "sv_ind_ptax_flag_w_low_price": "Low price",
+        "sv_ind_price_high_price_sqft": "High price per square foot",
+        "sv_ind_ptax_flag_w_high_price_sqft": "High price per square foot",
+        "sv_ind_price_low_price_sqft": "Low price per square foot",
+        "sv_ind_ptax_flag_w_low_price_sqft": "Low price per square foot",
```
wagnerlmichael, line 180: Added new ptax cols

glue/sales_val_flagging.py
```diff
@@ -152,30 +156,119 @@ def group_size_adjustment(df, stat_groups: list, min_threshold, condos: bool):
...
+        "sv_ind_price_high_price_sqft",
+        "sv_ind_price_low_price_sqft",
+    ]
 
-    # Using .loc[] to set the desired values for rows meeting the condition
-    merged_df.loc[condition, "sv_outlier_type"] = "Not outlier"
-    merged_df.loc[condition, "sv_is_outlier"] = 0
+    def fill_outlier_reasons(row):
+        reason_idx = 1
+        reasons_added = set()  # Set to track reasons already added
```
wagnerlmichael, line 212: The set allows functional iteration over two high price values in the dict

glue/sales_val_flagging.py
```diff
@@ -102,45 +102,49 @@ def ptax_adjustment(df, groups, ptax_sd, condos: bool):
     group_string = "_".join(groups)
 
     if not condos:
-        df["ptax_flag_w_deviation"] = df["ptax_flag_original"] & (
+        df["sv_ind_ptax_flag_w_high_price"] = df["ptax_flag_original"] & (
```
jeancochrane, line 105: **[Praise]** Nice work on this refactor!

glue/sales_val_flagging.py
```diff
@@ -152,30 +156,115 @@ def group_size_adjustment(df, stat_groups: list, min_threshold, condos: bool):
...
+        "sv_ind_price_low_price",
+        "sv_ind_price_high_price_sqft",
+        "sv_ind_price_low_price_sqft",
+    ]
 
-    # Using .loc[] to set the desired values for rows meeting the condition
-    merged_df.loc[condition, "sv_outlier_type"] = "Not outlier"
-    merged_df.loc[condition, "sv_is_outlier"] = 0
+    def fill_outlier_reasons(row):
+        reason_idx = 1
```
jeancochrane, line 211: **[Nitpick, non-blocking]** Now that we're keeping track of the set of `reasons_added`, I don't think we need this counter anymore -- we can just replace it with references to `len(reasons_added)`. Note that we'd need to be careful with off-by-one errors the context of the `outlier_types_crosswalk` loop, since we want to set the column name to `sv_outlier_reason{len(reasons_added) + 1}` and then subsequently check `if len(reasons_added) >= 3`. (We could also consider just setting `reason_idx = len(reasons_added) + 1` before setting `sv_outlier_reason{reason_idx}` in the loop if the `if` conditional passes.)

glue/sales_val_flagging.py
```diff
@@ -152,30 +156,115 @@ def group_size_adjustment(df, stat_groups: list, min_threshold, condos: bool):
...
+        reasons_added = set()  # Set to track reasons already added
+
+        for reason in outlier_type_crosswalk:
+            current_reason = outlier_type_crosswalk[reason]
```
jeancochrane, line 215: **[Nitpick, non-blocking]** Two tiny readability improvements you could make here:

1. It may not be immediately clear to a reader how `current_reason` is supposed to differ from `reason`; perhaps `reason_label` would be a clearer name for the variable?
2. Since we're iterating the keys and values of the crosswalk, we can just loop over the crosswalk's `items()`:

```python
for reason, reason_label in outlier_type_crosswalk.items():
    ...
```

glue/sales_val_flagging.py
```diff
@@ -152,30 +156,115 @@ def group_size_adjustment(df, stat_groups: list, min_threshold, condos: bool):
...
+            # Add a check to ensure that only specific reasons are added if _merge is not 'both'
+            if (
+                reason in row
+                and row[reason]
+                and current_reason
+                not in reasons_added  # Check if the reason is already added
+                # Apply group threshold logic
+                and not (row["_merge"] == "both" and reason in group_thresh_price_fix)
```
jeancochrane, line 223: **[Question, non-blocking]** I don't totally get how this line works, can you break it down for me?

manual_flagging/yaml/inputs.yaml
```diff
@@ -13,7 +13,7 @@ run_note: |
 # without flags, it assigns a version number of "1". The latest
 # version of each sale flag can be found in the "default"."vw_pin_sale" view.
 
-manual_update: true
+manual_update: false
```
jeancochrane, line 16: **[Question, non-blocking]** Do we want to persist these changes to `inputs.yml`, or were these just for testing purposes?

jeancochrane (approved):
Beautifully done! A couple of small nitpicks to simplify the code even further, but I think this is basically ready to go.

manual_flagging/yaml/inputs.yaml
wagnerlmichael: Ah, this was just for testing, I will revert.

glue/sales_val_flagging.py
wagnerlmichael: ### Condition 1  
`reason_ind_col in row`: this checks if the column exists in the dataset. This check is necessary because the sqft columns won't exist in the condos data

### Condition 2
`and row[reason_ind_col]`: this is the check for whether the ind col is true

### Condition 3

This checks to make sure that the column doesn't already exist in the `reasons_added_set`. Given that it is a set, it is kind of a redundant operation, but does it make sense to keep this for computation reduction? 

```python
and current_reason
not in reasons_added
```

Also I think it acts as a good check for `row[f"sv_outlier_reason{len(reasons_added) + 1}"] = current_reason` to make sure the column doesn't get populated with an incorrect reason name.


### Condition 4

`and not (row["_merge"] == "both" and reason_ind_col in group_thresh_price_fix`

This condition makes sure a price value won't be assigned to an obs with a group of N < `group_thresh`

jeancochrane: Thanks for these responses!

> [Condition 3] Given that it is a set, it is kind of a redundant operation, but does it make sense to keep this for computation reduction?

I think it makes sense to keep it -- the set will automatically deduplicate indicator values, but we still want to make sure we don't add duplicate values to `row[f"sv_outlier_reason{len(reasons_added) + 1}"]`.

> [Condition 4] This condition makes sure a price value won't be assigned to an obs with a group of `N < group_thresh`

Got it, that makes sense! Do you mind if we adjust the comment to make this a little bit clearer?

```python
            if (
                reason_ind_col in row
                and row[reason_ind_col]
                and current_reason
                not in reasons_added  # Check if the reason is already added
                # Apply group threshold logic: `row["_merge"]` will be `both` when the group threshold
                # is met, but only price indicators (`group_thresh_price_fix`) should use this threshold,
                # since ptax indicators have thresholds built-in
                and not (row["_merge"] == "both" and reason_ind_col in group_thresh_price_fix)
            ):
```

PR #129, opened 2024-05-31
Fix threshold outlier types

wagnerlmichael:
This PR addresses an issue in the `get_group_threshold()` function within the sales validation pipeline. Previously, we were setting the `sv_outlier_type` values to `Not outlier` only for certain values of `sv_outlier_type`.  

When we run the statistical flagging within a certain group that was below our `min_groups_threshold` defined in `inputs.yaml`, we should ideally be setting all non-ptax outliers to `Not outlier`. However, this was only happening for certain `sv_outlier_type` values. This resulted in some observations incorrectly being flagged as outliers when the number of observations in the group they belonged to was below the `min_group_threshold`.

The fix ensures that all observations in such groups are now consistently marked as `Not outlier`. This excludes ptax outliers, for which we forego the group threshold.

### DIFF
--- a/glue/sales_val_flagging.py
+++ b/glue/sales_val_flagging.py
@@ -160,26 +160,8 @@ def group_size_adjustment(df, stat_groups: list, min_threshold, condos: bool):
         df, filtered_groups[stat_groups], on=stat_groups, how="left", indicator=True
     )
 
-    # List of sv_outlier_type values to check
-    if condos == False:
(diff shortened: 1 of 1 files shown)


jeancochrane (approved):
Woof, thanks for catching this!

dfsnow (commented):
Nice catch @wagnerlmichael. Do you remember what the original intent of this conditional was? I can't recall

PR #130, opened 2024-06-10
Migrate data based on 2024 Q2 Sales Val re-design

wagnerlmichael:
This PR includes a script that will process our existing flags and output them into our sales val back up bucket. From there we will be able to migrate the new flags into the prod flag s3 bucket. The new flags and the old flags are both backed up in the backup bucket.

We take the existing flag table and replace `sv_outlier_type` with the new set of columns `sv_outlier_reason1`,  `sv_outlier_reason2`,  and `sv_outlier_reason3`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,11 +54,11 @@ Commercial, industrial, and land-only property sales are _not_ flagged by this m
 Outlier flags are broken out into 2 types: statistical outliers and heuristic outliers.
 
 - Statistical outliers are sales that are a set number of standard deviations (usually 2) away from the mean of a group of similar properties (e.g. within the same township, class, timeframe, etc.).
-- Heuristic outliers use some sort of existing flag or rule to identify potentially non-arms-length sales. Heuristic outliers are ***always combined with a statistical threshold***, i.e. a sale with a matching last name must _also_ be N standard deviations from a group mean in order to be flagged. Examples of heuristic outlier flags include:
-    - **PTAX flag**: The [PTAX-203](https://tax.illinois.gov/content/dam/soi/en/web/tax/localgovernments/property/documents/ptax-203.pdf) form is required by the Illinois Department of Revenue for most property transfers. Certain fields on this form are highly indicative of a non-arms-length transaction, i.e. Question 10 indicating a short sale.
-    - **Non-person sale**: Flagged keyword suggests the sale involves a non-person legal entity (industrial buyer, bank, real estate firm, construction, etc.).
-    - **Flip Sale**: Flagged when the owner of the home owned the property for less than 1 year
-    - **Anomaly**: Flagged via an unsupervised machine learning model (isolation forest).
+- Heuristic outliers use some sort of existing flag or rule to identify potentially non-arms-length sales. Heuristic outliers are _**always combined with a statistical threshold**_, i.e. a sale with a matching last name must _also_ be N standard deviations from a group mean in order to be flagged. Examples of heuristic outlier flags include:
+  - **PTAX flag**: The [PTAX-203](https://tax.illinois.gov/content/dam/soi/en/web/tax/localgovernments/property/documents/ptax-203.pdf) form is required by the Illinois Department of Revenue for most property transfers. Certain fields on this form are highly indicative of a non-arms-length transaction, i.e. Question 10 indicating a short sale.
+  - **Non-person sale**: Flagged keyword suggests the sale involves a non-person legal entity (industrial buyer, bank, real estate firm, construction, etc.).
+  - **Flip Sale**: Flagged when the owner of the home owned the property for less than 1 year
+  - **Anomaly**: Flagged via an unsupervised machine learning model (isolation forest).
 
 The following is a list of all current flag types:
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -123,54 +105,8 @@ As of 2024-03-15, around **6.9%** of the total sales have some sort of outlier c
...
-    default.vw_pin_sale
-CROSS JOIN
-    TotalRecords
-WHERE
-    sv_is_outlier IS NOT null
-    AND sv_outlier_type <> 'Not outlier'
-GROUP BY
-    sv_outlier_type, total_count
-ORDER BY
-    proportion DESC;
```
wagnerlmichael, line 147: Will need to re-work these queries. I was thinking of doing this post-migration

README.md
```diff
@@ -123,54 +105,8 @@ As of 2024-03-15, around **6.9%** of the total sales have some sort of outlier c
...
-|Anomaly (low)          |1.76%     |
-|High price (raw & sqft)|1.44%     |
-|Home flip sale (low)   |1.26%     |
-|Family sale (low)      |0.64%     |
-|Family sale (high)     |0.06%     |
-|High price swing       |0.01%     |
-|Low price swing        |0.00%     |
-
-
-*These outliers are flagged if the relevant price columns (log10 transformed and normalized) are 2 standard deviations below or above the mean within a given group*
```
wagnerlmichael, line 172: This will need to change as well.

migrations/0002_update_outlier_column_structure_w_iasworld_2024_update.py
```diff
@@ -0,0 +1,175 @@
...
+"""
+
+recode_dict = {
+    "PTAX-203 flag (Low)": {
+        "sv_outlier_reason1": "PTAX-203 Exclusion",
+        "sv_outlier_reason2": np.nan,
+    },
+    "PTAX-203 flag (High)": {
+        "sv_outlier_reason1": "PTAX-203 Exclusion",
+        "sv_outlier_reason2": np.nan,
```
wagnerlmichael, line 103: Split these into their component parts according to our new system.

migrations/0002_update_outlier_column_structure_w_iasworld_2024_update.py
```diff
@@ -0,0 +1,173 @@
...
+        df.insert(pos, f"sv_outlier_reason{i}", np.nan)
+        pos += 1
+
+    print(df.dtypes)
+    # Use the dictionary to populate the new columns
+    for key, value in recode_dict.items():
+        mask = df["sv_outlier_type"] == key
+        for col, val in value.items():
+            df.loc[mask, col] = val
+    print(df.dtypes)
```
jeancochrane, line 67: **[Nitpick, non-blocking]** Let's strip out these stray `print` statements:

```suggestion
    # Use the dictionary to populate the new columns
    for key, value in recode_dict.items():
        mask = df["sv_outlier_type"] == key
        for col, val in value.items():
            df.loc[mask, col] = val
```

migrations/0002_update_outlier_column_structure_w_iasworld_2024_update.py
jeancochrane: **[Question, non-blocking]** Remind me why we're getting rid of the price information for PTAX-flagged sales? I would expect this to be `sv_outlier_reason1="PTAX-203 Exclusion"` and `sv_outlier_reason2="High price"`.

migrations/0002_update_outlier_column_structure_w_iasworld_2024_update.py
```diff
@@ -0,0 +1,173 @@
...
+    },
+    "Low price (sqft)": {
+        "sv_outlier_reason1": "Low price per square foot",
+        "sv_outlier_reason2": np.nan,
+    },
+    "Home flip sale (high)": {
+        "sv_outlier_reason1": "High price",
+        "sv_outlier_reason2": "Price swing / Home flip",
+    },
+    "Home flipe sale (low)": {
```
jeancochrane, line 147: **[Nitpick, blocking]** This typo might account for the missing reasons I noted in the `data-architecture` PR:

```suggestion
    "Home flip sale (low)": {
```

jeancochrane (approved):
Looking great! One small blocking typo, but this should be good to go once that's resolved.

wagnerlmichael:
https://github.com/ccao-data/model-sales-val/pull/130#discussion_r1638973401. We are getting rid of the price information because it is a different scale than our regular price outliers. We use 1 or 1.5 std instead of the uniform 2 for the rest of the price types. We also don't subject it to the group threshold requirement. PTAX has always been kind of its own thing, this is why I don't split it into 2. You'll have to let me know if that makes sense to you

migrations/0002_update_outlier_column_structure_w_iasworld_2024_update.py
wagnerlmichael: https://github.com/ccao-data/model-sales-val/pull/130#discussion_r1638973401. We are getting rid of the price information because it is a different scale than our regular price outliers. We use 1 or 1.5 std instead of the uniform 2 for the rest of the price types. We also don't subject it to the group threshold requirement. PTAX has always been kind of its own thing, this is why I don't split it into 2. You'll have to let me know if that makes sense to you

migrations/0002_update_outlier_column_structure_w_iasworld_2024_update.py
```diff
@@ -0,0 +1,171 @@
...
+
+recode_dict = {
+    "PTAX-203 flag (Low)": {
+        "sv_outlier_reason1": "Low price",
+        "sv_outlier_reason2": "PTAX-203 Exclusion",
+    },
+    "PTAX-203 flag (High)": {
+        "sv_outlier_reason1": "High price",
+        "sv_outlier_reason2": "PTAX-203 Exclusion",
+    },
```
wagnerlmichael, line 100: Switched these based on new ptax handling

jeancochrane (approved):
Looks great!

Issue #126, opened 2024-05-23
Migrate prod data to use new reason codes


jeancochrane:
Once https://github.com/ccao-data/model-sales-val/issues/125 is complete, we should migrate our prod flags to use the new scheme for reason codes so that our data is consistent. We can use the `ccao-data-warehouse-sale-backup-us-east-1` again to back up the old flags before the migration.

wagnerlmichael:
Completed here https://github.com/ccao-data/model-sales-val/commit/8674745df4e9e8da0edef6bc5ab1e7e0fbbc3779

wagnerlmichael:
Completed here https://github.com/ccao-data/model-sales-val/commit/8674745df4e9e8da0edef6bc5ab1e7e0fbbc3779

PR #132, opened 2024-07-30
Refactor the export script output to match new reason code schema

### LINKED ISSUE #131: Refactor the export script output to match new reason code schema
In #124, #125, and #126, we refactored the pipeline script to output a new format for reason codes. We should also update the `scripts/export.py` script to match these changes.

wagnerlmichael:
This PR updates our ias upload script based on our recent sales val re-work (#124, #125, #126) .

Closes #131

### DIFF
--- a/scripts/export.py
+++ b/scripts/export.py
@@ -28,11 +28,11 @@ OUTLIER_TYPE_CODES = {
     "High price": "2",
     "Low price per square foot": "3",
     "High price per square foot": "4",
-    "Non-person sale": "7",
-    "PTAX-203 Exclusion": "8",
-    "Statistical Anomaly": "9",
-    "Price swing / Home flip": "10",
-    "Family sale": "11",
+    "Non-person sale": "5",
+    "PTAX-203 Exclusion": "6",
+    "Statistical Anomaly": "7",
+    "Price swing / Home flip": "8",
+    "Family sale": "9",
 }
 
 if __name__ == "__main__":
--- a/scripts/export.py
+++ b/scripts/export.py
@@ -23,7 +23,6 @@ OUTLIER_REASON3_FIELD = "TODO2"
 RUN_ID_FIELD = "USER28"
 
 OUTLIER_TYPE_CODES = {
-    None: "0",
     "Low price": "1",
     "High price": "2",
     "Low price per square foot": "3",
--- a/scripts/export.py
+++ b/scripts/export.py
@@ -23,13 +23,11 @@ OUTLIER_REASON3_FIELD = "TODO2"
 RUN_ID_FIELD = "USER28"
 
 OUTLIER_TYPE_CODES = {
-    "": "0",
+    None: "0",
     "Low price": "1",
     "High price": "2",
     "Low price per square foot": "3",
     "High price per square foot": "4",
-    "High price per square foot": "5",
-    "Low price per square foot": "6",
     "Non-person sale": "7",
     "PTAX-203 Exclusion": "8",
     "Statistical Anomaly": "9",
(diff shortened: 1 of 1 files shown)


scripts/export.py
```diff
@@ -17,29 +17,22 @@
...
-    "Home flip sale (low)": "10",
-    "Low price (raw & sqft)": "11",
-    "Low price (raw)": "12",
-    "Low price (sqft)": "13",
-    "Low price swing": "14",
-    "Non-person sale (high)": "15",
-    "Non-person sale (low)": "16",
-    "PTAX-203 flag (High)": "17",
-    "PTAX-203 flag (Low)": "18",
+    None: "0",
```
wagnerlmichael, line 26: In the new spec we switched `"Not outlier"` to an NA value.

scripts/export.py
```diff
@@ -17,29 +17,22 @@
...
+    "Low price": "1",
+    "High price": "2",
+    "Low price per square foot": "3",
+    "High price per square foot": "4",
+    "Non-person sale": "5",
+    "PTAX-203 Exclusion": "6",
+    "Statistical Anomaly": "7",
+    "Price swing / Home flip": "8",
+    "Family sale": "9",
```
wagnerlmichael, line 35: There are now fewer values overall due to decoupling.

scripts/export.py
```diff
@@ -114,25 +113,29 @@
...
+    mask_all_not = ~flag_df[columns_to_check].isin(values_to_check).any(axis=1)
+    mask_any = flag_df[columns_to_check].isin(values_to_check).any(axis=1)
+
+    assert (flag_df[IS_OUTLIER_FIELD] == "N").equals(
+        mask_all_not
+    ), "If there is a price reason in one of the outlier reason fields, it should be classified as an outlier"
+
+    assert (flag_df[IS_OUTLIER_FIELD] == "Y").equals(
+        mask_any
+    ), "If there is no price reason in any of the outlier reason fields, it should not be an outlier"
```
wagnerlmichael, line 138: Some old tests no longer made sense so I added some new ones.

scripts/export.py
```diff
@@ -114,25 +113,29 @@
...
-        f"{OUTLIER_TYPE_FIELD} must be {OUTLIER_TYPE_CODES['Not outlier']} "
-        f"when {IS_OUTLIER_FIELD} is N"
-    )
+    for field in [OUTLIER_REASON1_FIELD, OUTLIER_REASON2_FIELD, OUTLIER_REASON3_FIELD]:
+        invalid_values = flag_df[~flag_df[field].isin(OUTLIER_TYPE_CODES.values())]
+        assert invalid_values.empty, f"{field} contains invalid codes"
+
+    # Tests confirming that a price outlier reason is needed
+    # for a sale to be counted as an outlier, and vice-versa
+    values_to_check = list(OUTLIER_TYPE_CODES.values())[1:5]
```
jeancochrane, line 122: **[Suggestion, non-blocking]** I think we'll thank ourselves later if we build this check slightly more defensively -- if we decide to change the order of the codes in the `OUTLIER_TYPE_CODES` dict, but we don't remember to update this line, we could end up changing the conditions of this check by accident. Slightly better might be to define a constant like so that could tie this check to the outlier codes themselves:

```python
PRICE_CODES = ["Low price", "High price", "Low price per square foot", "High price per square foot"]

...

mask_all_not = ~flag_df[columns_to_check].isin(PRICE_CODES).any(axis=1)
mask_any = flag_df[columns_to_check].isin(PRICE_CODES).any(axis=1)
```

It's possible that we'll change the outlier codes in the future, but I expect that if/when we do so we'll be feeling more sensitive to the possible side effects of the change than if we reordered the `OUTLIER_TYPE_CODES` dict. We could also add a quick check to alert us of this case:

```python
for price_code in PRICE_CODES:
    assert price_code in OUTLIER_TYPE_CODES, f"{price_code} is in PRICE_CODES but missing from OUTLIER_TYPE_CODES"
```

scripts/export.py
jeancochrane: **[Suggestion, non-blocking]** I think in this case we will most likely actually want to leave nulls as nulls, since iasworld will also expect nulls in this case (and not a code like `0`). I'll make a note to confirm this with Mirella, but in the meantime let's switch to leaving nulls as-is and we can adjust later if need be.

jeancochrane (approved):
Awesome work! Are you down to run a fresh export and send me the file so I can spot check it? I left two small suggestions below, but otherwise this looks great to me!

scripts/export.py
```diff
@@ -17,29 +17,22 @@
...
-    "Home flip sale (low)": "10",
-    "Low price (raw & sqft)": "11",
-    "Low price (raw)": "12",
-    "Low price (sqft)": "13",
-    "Low price swing": "14",
-    "Non-person sale (high)": "15",
-    "Non-person sale (low)": "16",
-    "PTAX-203 flag (High)": "17",
-    "PTAX-203 flag (Low)": "18",
+    None: None,
```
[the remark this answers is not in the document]
jeancochrane: I think this is probably fine!

jeancochrane (approved):
Looks great, thanks for doing the test export!

PR #133, opened 2024-08-28
Preliminary results on north tri outliers

wagnerlmichael:
Draft. Don't close without consulting due to dev tables being important for other EDA/modeling issues


wagnerlmichael:
Closed, we got all we needed from this to move forward with changes for 2025 modeling.

PR #134, opened 2024-10-04
Fix `is_ptax_outlier` column

wagnerlmichael:
This PR corrects the `sv_is_ptax_outlier` column. It currently has all false values which is incorrect. We were calculating this incorrectly in [previous iterations](s3://ccao-data-backup-us-east-1/0003_fix_is_ptax_outlier_column/new_prod_data/flag/), specifically in the `finish_flags` and ptax_adjustment` functions of the sales val pipeline. However with our [most recent re-work](https://github.com/ccao-data/model-sales-val/pull/128) it functions correctly.

### DIFF
--- a/migrations/0003_fix_is_ptax_outlier_col.py
+++ b/migrations/0003_fix_is_ptax_outlier_col.py
@@ -1,3 +1,7 @@
+# Migration script to recompute the `sv_is_ptax_outlier` column for all
+# existing sales val data. We needed to do this because an earlier version
+# of the sales val pipeline accidentally used the wrong reason codes to
+# compute this column, resulting in every row getting set to False
 import os
 import awswrangler as wr
 import pandas as pd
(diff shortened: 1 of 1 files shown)


migrations/0003_fix_is_ptax_outlier_col.py
```diff
@@ -0,0 +1,102 @@
+import os
```
jeancochrane, line 1: **[Suggestion, non-blocking]** How about we add a brief docstring explaining what this migration is for, so that we can quickly remind ourselves in the future?

```suggestion
# Migration script to recompute the `sv_is_ptax_outlier` column for all
# existing sales val data. We needed to do this because an earlier version
# of the sales val pipeline accidentally used the wrong reason codes to
# compute this column, resulting in every row getting set to False
import os
```

jeancochrane (approved):
Thanks for catching this so fast!

migrations/0003_fix_is_ptax_outlier_col.py
wagnerlmichael: Oh yeah good call, thanks!

PR #135, opened 2024-12-06
Add unseen sale flag functionality

wagnerlmichael:


manual_flagging/flagging.py
```diff
@@ -423,13 +427,18 @@ def create_bins_and_labels(input_list):
...
 
-# Write to sale.group_mean table
+run_type = (
+    "initial_flagging"
+    if not inputs["manual_update"]
+    else "manual_update_only_new_sales"
+    if inputs["manual_update_only_new_sales"]
+    else "manual_update"
+)
+
```
wagnerlmichael, line 437: I think this works well without any need for a migration or meta data restructure.

manual_flagging/yaml/inputs.yaml
```diff
@@ -13,7 +13,15 @@ run_note: |
 # without flags, it assigns a version number of "1". The latest
 # version of each sale flag can be found in the "default"."vw_pin_sale" view.
 
-manual_update: true
+manual_update: false
```
jeancochrane, line 16: **[Nitpick, non-blocking]** While we're here, can we remove the stray space that separates this key from the comment above it 😅

manual_flagging/yaml/inputs.yaml
```diff
@@ -13,7 +13,15 @@ run_note: |
...
+
+# manual_update_only_new_sales: A boolean that only is in play when
+# manual_update = true. If true, the sales val pipeline will only
+# flag sale that currently have no existing flag in the sale.flag
+# table. In other words, it will flag unseen sales, not sales that
+# the sales val model has already given a flag value to.
+
+manual_update_only_new_sales: true
```
jeancochrane, line 24: **[Nitpick, non-blocking]** Same here! Plus a couple typos:

```suggestion
# manual_update_only_new_sales: A boolean that only is in play when
# manual_update = true. If true, the sales val pipeline will only
# flag sales that currently have no existing flag in the sale.flag
# table. In other words, it will flag unseen sales, not sales that
# the sales val model has already given a flag value to.
manual_update_only_new_sales: true
```

jeancochrane (changes requested):
Nice work doing this with so few changes! I think this is a good incremental change, although sometime next year we should take a second pass at this whole script and see if we can rewrite it with a simpler interface.

To confirm, we could probably approximate this behavior without the change by setting `time_frame` to only encompass the last year of sales, correct? I support this change regardless, but I want to make sure I understand whether it's mission-critical or a nice-to-have at this stage.

Do you mind also tweaking the [README section for model run modes](https://github.com/ccao-data/model-sales-val?tab=readme-ov-file#model-run-modes) to explain this case? I think it probably makes sense to remove the bullet for "Glue Job" (since we don't use that anymore), replace it with "Manual Update", and then add a third bullet for "Manual Update (New Sales Only)". It would also be helpful to explain the combination of the `manual_update` and `manual_update_only_new_sales` configs that produces each case, perhaps in a table.

My only blocking concern is the stray removal of the `condos` config, which I suspect should be done in a follow-up PR. Give me a quick re-request once you've responded to that and I'll approve!

manual_flagging/yaml/inputs.yaml
```diff
@@ -24,7 +32,6 @@ housing_market_type: [
   "res_single_family",
   "res_multi_family",
   "res_all",
-  "condos"
```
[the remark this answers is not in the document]
wagnerlmichael: Yep, unrelated

wagnerlmichael:
Hm, that is a good point about isolating the `time_frame` to approximate the change. What we would lose in that case are any sales that have been added prior to 2024 that we haven't flagged. I'm guessing this isn't very common, but it's a point of consideration. What do you think about just scrapping this PR and using time_frame, and lumping this into a big refactor later vs. implementing this change now? 

Update: I actually looked to see if we had added sales for prior years by comparing the flag data in `vw_pin_sale` with the ingest in sales val. 2022 and back it seems it's about the same. However it seems like about 2000 sales were retroactively added in 2023, or at least we haven't flagged them. I think this point supports the implementation of this PR.

jeancochrane:
The current situation based on my sync call with @wagnerlmichael:

* We have some new sales in ias for years 2016-2023, that were added retroactively since we last ran a flagging job. Most of these sales are in 2023, and the majority are in November/December 2023.
* We have some sales in ias for early 2024, but not many.
* We have extra mydec sales for all years, since the ias ingest from mydec performs some sale filtering.
* We also have some sales in mydec for 2024 that will _eventually_ get loaded into ias, but are not in ias right now.

Here's our plan:

1. We're going to run a flagging job on the combined mydec/ias view and only persist flags for unseen sales from 2023 to present.
2. Once 2024 sales are fully loaded into ias, we will run a flagging job on ias sales only and only persist flags for all sales from 2023 to present.
3. Once we've flagged all 2024 sales that are in ias, we will drop any flags from `sale.flag` for sales that are in mydec but not in ias.
4. Finally, we will run a flagging job on the ias view on all unseen sales prior to 2023, to catch new sales that were retroactively added this year.

Note that this workflow means that there will be some 2023 sales that we already flagged last year but whose flags will update for this year. Michael and I think this is fine, since we can still back out the version of the flag that we used for the 2024 model by filtering for flagging run dates that were before the model run date and then sorting the flags by version.

README.md
```diff
@@ -137,21 +139,19 @@ graph TD
...
-        G2[Save results to S3 with unique run ID]
-
-        A2 -->|Trigger| B2
-        B2 -->|Process new sales| C2
-        C2 -->|Run model| D2
-        D2 -->|Output flags| E2
-        E2 -->|Join data| F2
-        F2 -->|Persist results| G2
+    subgraph "Manual Update (New Sales Only) Mode"
+        A4{{"Flag only new sales"}}
```
jeancochrane, line 143: **[Nitpick, blocking]** Quick change: Looks like we need to swap out the `style B2` block on line 171 so that it says `style A4` and highlights this node, right?

wagnerlmichael: Right, good catch

jeancochrane (approved):
Nice work!

PR #138, opened 2024-12-13
Fix rename error and add docs

wagnerlmichael:
This has been a persistent weird error that flips back and forth between which line of code works. It has happened since the inception of the sales val pipeline. I'm not exactly sure why this happens, but it has always worked with one of the two options in the comment.

### DIFF
--- a/glue/flagging_script_glue/flagging.py
+++ b/glue/flagging_script_glue/flagging.py
@@ -575,7 +575,7 @@ def get_sale_counts(dups: pd.DataFrame) -> pd.DataFrame:
         dups.pin.value_counts().reset_index()
         # NOTE: For some unknown reason, using either of the following rename methods
         # may cause unexpected issues in certain environments. If one option breaks the code,
-        # try the usually works:
+        # the other usually works:
         #
         # .rename(columns={'count':'sv_sale_dup_counts'})
         # .rename(columns={'index':'pin', 'pin':'sv_sale_dup_counts'})
--- a/glue/flagging_script_glue/flagging.py
+++ b/glue/flagging_script_glue/flagging.py
@@ -572,9 +572,14 @@ def get_sale_counts(dups: pd.DataFrame) -> pd.DataFrame:
         df (pd.DataFrame): pandas dataframe
     """
     v_counts = (
-        dups.pin.value_counts()
-        .reset_index()
-        .rename(columns={"count": "sv_sale_dup_counts"})
+        dups.pin.value_counts().reset_index()
+        # NOTE: For some unknown reason, using either of the following rename methods
+        # may cause unexpected issues in certain environments. If one option breaks the code,
+        # try the usually works:
+        #
+        # .rename(columns={'count':'sv_sale_dup_counts'})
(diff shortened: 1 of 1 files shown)


glue/flagging_script_glue/flagging.py
```diff
@@ -572,9 +572,14 @@ def get_sale_counts(dups: pd.DataFrame) -> pd.DataFrame:
...
-        .reset_index()
-        .rename(columns={"count": "sv_sale_dup_counts"})
+        dups.pin.value_counts().reset_index()
+        # NOTE: For some unknown reason, using either of the following rename methods
+        # may cause unexpected issues in certain environments. If one option breaks the code,
+        # the other usually works:
+        #
+        # .rename(columns={'count':'sv_sale_dup_counts'})
+        # .rename(columns={'index':'pin', 'pin':'sv_sale_dup_counts'})
+        .rename(columns={"index": "pin", "pin": "sv_sale_dup_counts"})
```
jeancochrane, line 582: **[Suggestion, non-blocking]** I suspect the problem here has to do with the difference in the way `value_counts()` names the output Series between pandas 1.3.5 (which is required in [`manual_flagging/requirements.txt`](https://github.com/ccao-data/model-sales-val/blob/d7e1b9c070ec6a401c9bb3845644c2af9dbdaeee/manual_flagging/requirements.txt#L14)) and in pandas 2.x.

Here's pandas 1.3.5:

```python
>>> pd.DataFrame({"pin": [3, 1, 2, 3, 4], "val": ["a", "b", "c", "d", "e"]}).pin.value_counts()
3    2
1    1
2    1
4    1
Name: pin, dtype: int64
```

Notice that the name of the non-index (count) column is `pin`. Here's pandas 2.2.3, by contrast:

```python
>>> pd.DataFrame({"pin": [3, 1, 2, 3, 4], "val": ["a", "b", "c", "d", "e"]}).pin.value_counts()
pin
3    2
1    1
2    1
4    1
Name: count, dtype: int64
```

Now the index has the name `pin`, and the count column is called `count`. Based on these two interfaces, I would expect the commented-out code on line 580 to work for pandas 2.2.3, while the code on line 581 works for pandas 1.3.5.

This block of code should work the same in both versions of pandas:

```suggestion
        dups.pin.value_counts().reset_index(name="count")
        .rename(columns={"index": "pin"})
```

In both versions, the `name="count"` kwarg to [`reset_index()`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.reset_index.html) will cause the count column to be called "count". Then in pandas 1.3.5 the `rename()` call changes the column named "index" to "pin", whereas in pandas 2.2.3 the `rename()` call is a no-op because the column is already named "pin" and there is no "index" column in the dataframe.

Does this sound right to you? Let me know if not and we can debug further! But I think a solution that addresses the root cause of the discrepancy and is portable between all versions of pandas is the right way to go here.

jeancochrane (approved):
One thought about how we might resolve the root cause of the problem below, but it's not a big deal either way!

PR #139, opened 2024-12-16
Reformat method chain to format properly

wagnerlmichael:
I merged https://github.com/ccao-data/model-sales-val/pull/138 too quickly and didn't let the pre-commit check run. Here is the fix.


jeancochrane (approved):
Looks good!

PR #140, opened 2024-12-16
Fix reference to `sv_sale_dup_counts`

wagnerlmichael:
This [column ref](https://github.com/ccao-data/model-sales-val/blob/faee8ec26c68f6431d740ce2ea20e04409fc69fe/manual_flagging/yaml/inputs.yaml#L100) was changed in https://github.com/ccao-data/model-sales-val/pull/138, they need to be consistent so I switched it back


jeancochrane (approved):
Oops, good catch!

PR #142, opened 2025-01-03
Add raw price threshold for sales val

wagnerlmichael:
In the current methodology, a sale with a price of 10 billion could make it into the model. This is because the statistical groups sales val uses to flag sales automatically include the sales if its group has less than 30 sales. We add a cap that automatically disqualifies sales regardless of the group threshold.

We also standardize instances of "thresh" naming to "threshold"

### DIFF
--- a/glue/flagging_script_glue/flagging.py
+++ b/glue/flagging_script_glue/flagging.py
@@ -807,6 +807,10 @@ def outlier_type(df: pd.DataFrame, condos: bool) -> pd.DataFrame:
             "sv_ind_price_low_price_sqft",
         ]
 
+    # Implement raw threshold
+    price_conditions.append(df["sale_price"] > 9_000_000)
+    price_labels.append("sv_raw_price_threshold")
+
     combined_conditions = price_conditions + char_conditions
     combined_labels = price_labels + char_labels
 
--- a/glue/flagging_script_glue/flagging.py
+++ b/glue/flagging_script_glue/flagging.py
@@ -808,7 +808,7 @@ def outlier_type(df: pd.DataFrame, condos: bool) -> pd.DataFrame:
         ]
 
     # Implement raw threshold, unlog  price
-    price_conditions.append((10 ** df["meta_sale_price"]) > 1_000_000)
+    price_conditions.append((10 ** df["meta_sale_price"]) > 15_000_000)
     price_labels.append("sv_ind_raw_price_threshold")
 
     combined_conditions = price_conditions + char_conditions
--- a/glue/flagging_script_glue/flagging.py
+++ b/glue/flagging_script_glue/flagging.py
@@ -807,9 +807,9 @@ def outlier_type(df: pd.DataFrame, condos: bool) -> pd.DataFrame:
             "sv_ind_price_low_price_sqft",
         ]
 
-    # Implement raw threshold
-    price_conditions.append(df["sale_price"] > 9_000_000)
-    price_labels.append("sv_raw_price_threshold")
+    # Implement raw threshold, unlog  price
+    price_conditions.append((10 ** df["meta_sale_price"]) > 1_000_000)
+    price_labels.append("sv_ind_raw_price_threshold")
 
     combined_conditions = price_conditions + char_conditions
     combined_labels = price_labels + char_labels
--- a/glue/flagging_script_glue/flagging.py
+++ b/glue/flagging_script_glue/flagging.py
@@ -21,6 +21,7 @@ def go(
     iso_forest_cols: list,
     dev_bounds: tuple,
     condos: bool,
+    raw_price_thresh: int,
 ):
     """
     This function runs all of our other functions in the correct sequence.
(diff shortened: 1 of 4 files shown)


glue/sales_val_flagging.py
```diff
@@ -178,6 +178,7 @@ def classify_outliers(df, stat_groups: list, min_threshold):
         "sv_ind_ptax_flag_w_high_price_sqft": "High price per square foot",
         "sv_ind_price_low_price_sqft": "Low price per square foot",
         "sv_ind_ptax_flag_w_low_price_sqft": "Low price per square foot",
+        "sv_ind_raw_price_threshold": "Raw price threshold",
```
wagnerlmichael, line 181: Assigning this reason with priority right after the other price flags that are generated with the statistical flagging groups.

Order of assignment between the three outlier reasons will be

- Price outlier
- Raw price outlier (this cap)
- Ptax
- Other char reasons

glue/flagging_script_glue/flagging.py
```diff
@@ -807,6 +807,10 @@ def outlier_type(df: pd.DataFrame, condos: bool) -> pd.DataFrame:
             "sv_ind_price_low_price_sqft",
         ]
 
+    # Implement raw threshold, unlog  price
+    price_conditions.append((10 ** df["meta_sale_price"]) > 15_000_000)
```
wagnerlmichael, line 811: This value is pending a final decision

manual_flagging/yaml/inputs.yaml
```diff
@@ -163,3 +163,7 @@ ptax_sd: [1, 1]
 
 # Flags are only applied if there are at least this many sales in the group
 min_groups_threshold: 30
+
+# This is the raw price threshold that is used to set sales to outlier status
+# regardless of group size
+raw_price_thresh: 1500000
```
jeancochrane, line 169: **[Nitpick, non-blocking]** We use the full word `threshold` in other variables like `min_groups_threshold`, let's keep that same pattern here:

```suggestion
raw_price_threshold: 1500000
```

glue/sales_val_flagging.py
```diff
@@ -199,6 +201,11 @@ def classify_outliers(df, stat_groups: list, min_threshold):
 
     Note: This doesn't apply for sales that also have a ptax outlier status.
           In this case, we still assign the price outlier status.
+
+          We also don't apply this threshold with sv_raw_price_threshold,
+          since this is design to be a safeguard that catches very high price
```
jeancochrane, line 206: **[Nitpick, non-blocking]** Small typo:

```suggestion
          since this is designed to be a safeguard that catches very high price
```

jeancochrane (approved):
Looks good! Do we also need to save this value to the `parameter` table as a new column?

glue/sales_val_flagging.py
```diff
@@ -523,9 +523,9 @@ def get_parameter_df(
         "ptax_sd": [ptax_sd],
         "rolling_window": [rolling_window],
         "time_frame": [time_frame],
-        "short_term_owner_threshold": [short_term_thresh],
-        "min_group_thresh": [min_group_thresh],
-        "raw_price_thresh": [raw_price_thresh],
+        "short_term_owner_threshold": [short_term_threshold],
+        "min_group_thresh": [min_group_threshold],
```
wagnerlmichael, line 527: This min_group_thresh should eventually be migrated to min_group_threshold to maintain naming style.

glue/sales_val_flagging.py
```diff
@@ -490,6 +500,7 @@ def get_parameter_df(
         date_floor: parameter specification that limits earliest flagging write
         short_term_thresh: short-term threshold for Mansueto's flagging model
         min_group_thresh: minimum group size threshold needed to flag as outlier
+        raw_price_thresh: raw price threshold at which we unconditionally classify sales as outliers
```
jeancochrane, line 503: **[Nitpick, non-blocking]** Two param names that need to be updated to match the new function interface:

```suggestion
        short_term_threshold: short-term threshold for Mansueto's flagging model
        min_group_thresh: minimum group size threshold needed to flag as outlier
        raw_price_threshold: raw price threshold at which we unconditionally classify sales as outliers
```

manual_flagging/yaml/inputs.yaml
```diff
@@ -163,3 +163,7 @@ ptax_sd: [1, 1]
 
 # Flags are only applied if there are at least this many sales in the group
 min_groups_threshold: 30
+
+# This is the raw price threshold that is used to set sales to outlier status
+# regardless of group size
+raw_price_threshold: 1500000
```
jeancochrane, line 169: **[Nitpick, non-blocking]** Unless my eyes deceive me, this is the wrong order of magnitude, right?

```suggestion
raw_price_threshold: 15_000_000
```

jeancochrane (approved):
Thanks for this! Just make sure that the threshold is correct before merging.

manual_flagging/yaml/inputs.yaml
wagnerlmichael: TIL yaml has the underscore spacing

PR #145, opened 2025-01-21
Fix keyword arg

wagnerlmichael:
We forgot to update a keyword arg when [we did some standardizing of naming](https://github.com/ccao-data/model-sales-val/commit/962ef1da010b9599a43755bf19bfd690c75b16c5).


jeancochrane (approved):
Oops, good catch!

PR #147, opened 2025-02-06
Refactor standard deviation calculation

wagnerlmichael:


glue/flagging_script_glue/flagging.py
```diff
@@ -186,238 +186,167 @@ def pricing_info(
...
     if not condos:
-        df["sv_which_price"] = df.apply(which_price, args=(holds, groups), axis=1)
-
+        col = f"sv_price_per_sqft_deviation_{group_str}"
+        df[f"{col}_lower"] = df.groupby(list(groups))[col].transform("mean") - permut[
+            0
+        ] * df.groupby(list(groups))[col].transform("std")
+        df[f"{col}_upper"] = df.groupby(list(groups))[col].transform("mean") + permut[
+            1
+        ] * df.groupby(list(groups))[col].transform("std")
```
jeancochrane, line 244: **[Suggestion, non-blocking]** Nice generalization! We could go one step further and fold the `if not condos` branch into the for loop that precedes it:

```python
thresh_cols = [
    f"sv_price_deviation_{group_str}",
    f"sv_cgdr_deviation_{group_str}"
] + ([] if condos else ["f"sv_price_per_sqft_deviation_{group_str}"])

for col in thresh_cols:
    df[f"{col}_lower"] = df.groupby(list(groups))[col].transform("mean") - permut[
        0
    ] * df.groupby(list(groups))[col].transform("std")
    ...
```

glue/flagging_script_glue/flagging.py
```diff
@@ -186,238 +186,167 @@ def pricing_info(
...
-                p_std, *p_std_range = thresholds.get(
-                    f"sv_cgdr_deviation_{group_string}"
-                ).get(key)
-
-                p_lower, p_upper = p_std_range
+        raw_out = raw_val > raw_upper or raw_val < raw_lower
+        sqft_val = row[f"sv_price_per_sqft_deviation_{group_str}"]
+        sqft_lower = row[f"sv_price_per_sqft_deviation_{group_str}_lower"]
+        sqft_upper = row[f"sv_price_per_sqft_deviation_{group_str}_upper"]
+        sqft_out = sqft_val > sqft_upper or sqft_val < sqft_lower
```
jeancochrane, line 331: **[Nitpick, non-blocking]** Any reason not to use `between_two_numbers()` here, the way we do for the rest of the threshold checks?

jeancochrane (approved):
I like this refactor! Can you share the query you ran to check that these results are identical to the existing results? Once I've verified that, we'll be good to go.

wagnerlmichael:
Reminder: There were inconsistencies between the flag outputs in main and this branch, so if we proceed with this work we'll need to figure that out

PR #160, opened 2025-09-16
Improve development workflow for sales val

wagnerlmichael:


README.md
```diff
@@ -6,6 +6,7 @@ Table of Contents
 - [Outlier Types](#outlier-types)
 - [Flagging Details](#flagging-details)
 - [Structure of Data](#structure-of-data)
+- [Developing the sales val pipeline](#developing-the-sales-val-pipeline)
```
jeancochrane, line 9: **[Nitpick, required]** Unfortunately due to the way GitHub linking works, the link here needs to match the header exactly:

```suggestion
- [Developing the sales val pipeline and testing changes](#developing-the-sales-val-pipeline-and-testing-changes)
```

README.md
```diff
@@ -245,6 +246,26 @@ erDiagram
     }
 ```
 
+## Developing the sales val pipeline, testing changes
```
jeancochrane, line 249: **[Nitpick, required]** Same here:

```suggestion
## Developing the sales val pipeline and testing changes
```

src/utils.py
```diff
@@ -572,18 +572,24 @@ def get_metadata_df(run_id, timestamp, run_type, commit_sha, run_note):
...
         run_id: unique run_id of the script
     """
+    USER = os.getenv("USER")
+
+    if output_environment == "prod":
+        base_path = os.path.join(os.getenv("AWS_S3_WAREHOUSE_BUCKET"), "sale")
+    elif output_environment == "dev":
+        base_path = os.path.join(
+            os.getenv("AWS_S3_WAREHOUSE_BUCKET_DEV"), f"z_dev_{USER}_sale"
+        )
```
jeancochrane, line 591: **[Suggestion, optional]** Some thoughts here:

- Since we've already validated that `output_environment` is one of "prod" or "dev", we can use an `else` clause
- No need for `os.path.join()` in this case because the path we're referring to is an S3 path, not an OS path, meaning it's guaranteed to use forward-slashes as separators
- Let's raise an error if `USER` is unset, instead of failing silently
- No need for the `AWS_WAREHOUSE_BUCKET` env vars, we can just hardcode them to make the script interface simpler

```suggestion
    if output_environment == "prod":
        base_path = "s3://ccao-data-warehouse-us-east-1/sale"
    else:
        USER = os.getenv("USER")
        if not USER:
            raise ValueError(
                "$USER environment variable is unset but is required when "
                "output_environment == 'dev'"
            )            
        base_path = f"s3://ccao-data-warehouse-us-east-1/z_dev_{USER}_sale"
```

src/utils.py
```diff
@@ -572,18 +572,24 @@ def get_metadata_df(run_id, timestamp, run_type, commit_sha, run_note):
...
+
     file_name = run_id + ".parquet"
-    s3_file_path = os.path.join(
-        s3_warehouse_bucket_path, "sale", table_name, file_name
-    )
+    s3_file_path = os.path.join(base_path, table_name, file_name)
```
jeancochrane, line 594: **[Nitpick, optional]** Same here:

```suggestion
    s3_file_path = f"{base_path}/{table_name}/{file_name}"
```

jeancochrane (approved):
Nice work!

Issue #152, opened 2025-09-11
Switch to using uv for Python environment management


jeancochrane:
We don't generally use `requirements.txt` files anymore, so let's switch to uv to match the standard pattern we follow in our other Python projects.

We also have a bunch of dependabot alerts, so let's take this opportunity to upgrade our dependencies to secure versions, too.

PR #164, opened 2025-11-05
Overhaul sales val pipeline with DVC, add uv

### LINKED ISSUE #161: Create versioned archive of input data
As we QC sales val flags over time, we're increasingly finding that it would be useful to be able to exactly replicate the input data that we used to generate a given set of flags. We should create a versioned archive of input data, similar to the way we archive input data in the res and condo models.

I would be happy to use [DVC](https://dvc.org/) for this task in order to more closely match the res and condo model approach to data versioning, but I also know we've found DVC to be burdensome in the past, so I'm open to other ideas too.

wagnerlmichael:
This PR adds DVC and switches from pip and `requirements.txt` to using uv.

Data written in this PR is sitting in `z_dev_miwagne_sale` Database.

## Change highlights

- We divide what was `flag.py` into `00_ingest.py`, `01_flag.py`, and `02_upload.py`. This allows a smoother dvc experience and truly independent stages
- Removed `requirements.txt` for `uv.lock` and beefed up `pyproject.toml`
- The pipeline and helper/model files all exist in `src/`. I thought about trying to make subdirectories but I didn't see a clear improvement. Open to suggestions
- Added `input/` and `output/` directories like we have in the res and condo model repos
- Fix a bug introduced by bumping packages in `src/model.py`

Closes #161 and https://github.com/ccao-data/model-sales-val/issues/152

### DIFF
--- a/dvc.lock
+++ b/dvc.lock
@@ -17,24 +17,24 @@ stages:
       size: 4624
     - path: src/utils.py
       hash: md5
-      md5: 075eb769fd7349744770bf6defc3bc3e
-      size: 21525
+      md5: 1334995fa8d941884d1a91e6608e3739
+      size: 21537
     outs:
     - path: input/sales_ingest.parquet
       hash: md5
-      md5: 6fbaf9eca1730771d738c738a92d1821
-      size: 9087367
+      md5: 013af04502243c8ebc367c204c3dbada
+      size: 9091686
   flag:
     cmd: uv run --env-file .env python3 src/01_flag.py
     deps:
     - path: input/sales_ingest.parquet
       hash: md5
-      md5: 6fbaf9eca1730771d738c738a92d1821
-      size: 9087367
+      md5: 013af04502243c8ebc367c204c3dbada
+      size: 9091686
     - path: src/01_flag.py
       hash: md5
-      md5: d09fab143bdc7ce5220d37100cdd7c99
-      size: 11785
+      md5: cabbfca21530fabc714d7d9d76c34f28
+      size: 11949
     - path: src/constants.py
       hash: md5
       md5: 7c69d793d7e5917f4d6413316e64166f
@@ -49,44 +49,44 @@ stages:
       size: 44748
     - path: src/utils.py
       hash: md5
-      md5: 075eb769fd7349744770bf6defc3bc3e
-      size: 21525
+      md5: 1334995fa8d941884d1a91e6608e3739
+      size: 21537
     outs:
     - path: output/flag.parquet
       hash: md5
-      md5: 606d850ef98b63508944009fb5ecc4be
-      size: 328774
+      md5: e52d6f26f0d1a89338b60cedb4d30b21
+      size: 328779
     - path: output/group_mean.parquet
       hash: md5
-      md5: 2f94e90d155515d588e70d81f44d4523
-      size: 39663
+      md5: 53c87becb06319e72b77f5702b8cc7e4
+      size: 39668
     - path: output/metadata.parquet
       hash: md5
-      md5: e20a40a32d456decd7d70de46255ad8c
-      size: 4226
+      md5: 7e704bb76e8e690b2960f37b83c30316
+      size: 4231
     - path: output/parameter.parquet
       hash: md5
-      md5: 393d8942cb2ce32eae5a9a42f9466ee8
-      size: 16200
+      md5: 797ee37c3f6ba00b9c047c00eb32bd38
+      size: 16205
   upload:
     cmd: uv run --env-file .env python3 src/02_upload.py
     deps:
     - path: output/flag.parquet
       hash: md5
-      md5: 606d850ef98b63508944009fb5ecc4be
-      size: 328774
+      md5: e52d6f26f0d1a89338b60cedb4d30b21
+      size: 328779
     - path: output/group_mean.parquet
       hash: md5
-      md5: 2f94e90d155515d588e70d81f44d4523
-      size: 39663
+      md5: 53c87becb06319e72b77f5702b8cc7e4
+      size: 39668
     - path: output/metadata.parquet
       hash: md5
-      md5: e20a40a32d456decd7d70de46255ad8c
-      size: 4226
+      md5: 7e704bb76e8e690b2960f37b83c30316
+      size: 4231
     - path: output/parameter.parquet
       hash: md5
-      md5: 393d8942cb2ce32eae5a9a42f9466ee8
-      size: 16200
+      md5: 797ee37c3f6ba00b9c047c00eb32bd38
+      size: 16205
     - path: src/02_upload.py
       hash: md5
       md5: 6d9d84513d3bb7bf9b8bf8e9481fef57
@@ -97,5 +97,5 @@ stages:
       size: 4624
     - path: src/utils.py
       hash: md5
-      md5: 075eb769fd7349744770bf6defc3bc3e
-      size: 21525
+      md5: 1334995fa8d941884d1a91e6608e3739
+      size: 21537


.dvc/config
```diff
@@ -0,0 +1,4 @@
+[core]
+    remote = storage
```
jeancochrane: **[Nitpick, optional]** I suggest we copy over the other settings from [the res model](https://github.com/ccao-data/model-res-avm/blob/master/.dvc/config), though I don't feel super strongly about it:

```suggestion
    analytics = false
    remote = storage
    autostage = true
```

Based on [the docs](https://dvc.org/doc/user-guide/project-structure/configuration#core), the effect of these changes will be:

- Disable anonymized usage analytics
- Automatically stage files that dvc changes

.dvc/config
```diff
@@ -0,0 +1,4 @@
...
+['remote "storage"']
+    url = s3://ccao-data-dvc-us-east-1/sales_val/
```
jeancochrane: **[Nitpick, required]** Let's use the exact repo name for this S3 prefix to make it as easy as possible for a future maintainer to trace it back to this repo.


```suggestion
    url = s3://ccao-data-dvc-us-east-1/model-sales-val/
```

pyproject.toml
```diff
@@ -0,0 +1,29 @@
+[project]
+name = "src"
+version = "0.0.1"
+readme = "README.md"
+requires-python = ">=3.11"
+# Note - this command opens ipython terminal using uv venv
+# uv run --with ipython ipython
```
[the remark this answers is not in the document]
jeancochrane: **[Praise]** I didn't know this particular incantation, I like it! Maybe this would be better placed in some sort of centralized documentation about uv, but since we don't really have that yet, I'm fine keeping it here.

jeancochrane (commented):
Awesome start! I tested out the ingest stage and it works great. I gave you permissions to push to the DVC remote, so give `dvc push` a try and see if it works for you. I haven't given the scripts a close read yet, but I'll wait until you've tested it to dive in.

src/00_ingest.py
```diff
@@ -0,0 +1,155 @@
...
+
+import yaml
+from pyathena import connect
+from pyathena.pandas.util import as_pandas
+
+import constants
+import utils
+
+root = sp.getoutput("git rev-parse --show-toplevel")
+os.chdir(os.path.join(root, "src"))
```
jeancochrane: **[Suggestion, optional]** I don't actually think we need to change into the `src/` subdirectory like this, and I think it introduces the opportunity for confusion during debugging, since a user who runs the script may not realize that it's changed to run in a different directory than their current working directory. I tested removing this block and adjusting all of the filepaths to be relative to the top level of the project, rather than relative to the `src/` subdirectory, and it seemed to work fine.

Is there any reason not to remove this block and change the filepaths in the script to assume that the user is running the script from the root directory of the project? In particular, I'm wondering if maybe there's something about the way you run this code in your editor that makes this block useful. I run the script in the terminal, but I get the impression that working directories can be a little bit more annoying when running from VSCode (or Positron, by extension).

src/02_upload.py
```diff
@@ -0,0 +1,44 @@
...
+root = sp.getoutput("git rev-parse --show-toplevel")
+os.chdir(os.path.join(root, "src"))
+
+# Use yaml as inputs
+with open("inputs.yaml", "r") as stream:
+    inputs = yaml.safe_load(stream)
+
+root = sp.getoutput("git rev-parse --show-toplevel")
+os.chdir(os.path.join(root, "src"))
+
```
jeancochrane: **[Nitpick, non-blocking]** This seems to be a duplicate of lines 9-10.

```suggestion
```

src/00_ingest.py
wagnerlmichael: I tried dropping `os.chdir(os.path.join(root, "src"))` and adjusting the file paths in the python files. I then ran `dvc repro` and ran into this error
```
Traceback (most recent call last):
  File "/home/miwagne/repos/model-sales-val/src/00_ingest.py", line 10, in <module>
    import src.constants as constants
ModuleNotFoundError: No module named 'src'
```

From what I can tell this is happening because when python calls a script it automatically sets `sys.path` as the dir where the file is located, so the call here is looking in `src/src/`.

I thought about adjusting all of the the calls in `dvc.yaml` to eliminate `src/`, but it feels like we probably want to be running `dvc repro` from the root of the repo. Does that sound right?

My best idea is to change it to a module setup `cmd: uv run --env-file .env python -m src.00_ingest`, because [apparently the `-m` flag looks](https://docs.python.org/3/using/cmdline.html#run-module) for a module name and then grabs the working directory instead of the file's directory itself. I'm going to guess that you know a lot more about this than me so I'm curious if you think the `-m` flag is the way to go or there is a better option here.

src/00_ingest.py
```diff
@@ -0,0 +1,155 @@
...
+
+# Use yaml as inputs
+with open("inputs.yaml", "r") as stream:
+    inputs = yaml.safe_load(stream)
```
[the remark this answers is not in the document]
wagnerlmichael: Oh nice, that makes sense to me

src/00_ingest.py
```diff
@@ -0,0 +1,148 @@
...
+import subprocess as sp
+import time
+
+from pyathena import connect
+from pyathena.pandas.util import as_pandas
+
+import constants
+import utils
+
+root = sp.getoutput("git rev-parse --show-toplevel")
```
jeancochrane: **[Suggestion, optional]** I don't think we need to figure out the project root anymore, now that the script is reorganized under the assumption that we'll run it from the root.


```suggestion
```

src/01_flag.py
```diff
@@ -1,191 +1,66 @@
...
 import model
 import utils
 
-# Set working dir to manual_update, standardize yaml and src locations
 root = sp.getoutput("git rev-parse --show-toplevel")
-os.chdir(os.path.join(root, "src"))
-
-# Use yaml as inputs
-with open("inputs.yaml", "r") as stream:
-    inputs = yaml.safe_load(stream)
```
jeancochrane: **[Suggestion, optional]** Same here:

```suggestion
```

src/01_flag.py
```diff
@@ -458,25 +335,21 @@ def create_bins_and_labels(input_list):
...
-for table, df in tables_to_write.items():
-    utils.write_to_table(
-        df=df,
-        table_name=table,
-        run_id=run_id,
-        output_environment=inputs["output_environment"],
-    )
-    print(f"{table} table successfully written")
+for filename, df in tables_to_write.items():
+    output_path = os.path.join(os.path.join(root, "output"), filename)
```
jeancochrane: **[Suggestion, optional]**


```suggestion
    output_path = os.path.join("output", filename)
```

src/01_flag.py
```diff
@@ -1,191 +1,66 @@
...
-    NOT data.pin_is_multicard
-    OR data.source_table = 'condo_char'
+assert len(constants.INPUTS["run_tri"]) == len(
+    set(constants.INPUTS["run_tri"])
+), "Duplicate values in 'run_tri'"
+
+# Ingest
+df = pd.read_parquet(os.path.join(root, "input", "sales_ingest.parquet"))
+df_ingest = pd.read_parquet(
+    os.path.join(root, "input", "sales_ingest.parquet")
```
jeancochrane: **[Suggestion, optional]** Same here. Also, do we need both of these dataframes? If so, we could just use `copy` to avoid reading the file twice:

```suggestion
df = pd.read_parquet(os.path.join("input", "sales_ingest.parquet"))
df_ingest = df.copy()
```

src/constants.py
```diff
@@ -1,7 +1,12 @@
+import yaml
```
jeancochrane: **[Suggestion, required]** To support my suggestion below:


```suggestion
import os

import yaml
```

src/01_flag.py
wagnerlmichael: I don't think we need the second one, I deleted it.

.env
```diff
@@ -0,0 +1,2 @@
+AWS_ATHENA_S3_STAGING_DIR=s3://ccao-athena-results-us-east-1/
+AWS_REGION=us-east-1
```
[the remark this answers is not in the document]
wagnerlmichael: Very nice! I found that using this I could get my `iPython` shell to work going from `uv run --with ipython ipython` to `uv run --env-file .env --with ipython ipython
`

jeancochrane (approved):
This is great! I'll feel a lot better knowing that we're versioning our input data.

Issue #168, opened 2025-11-19
Persist dvc hash for input data


wagnerlmichael:
When we [updated the sales val pipeline](https://github.com/ccao-data/model-sales-val/pull/164) to use dvc and version the input data we forgot a key component.  We need to persist the input data md5 hash used in the dvc lockfile in order to be able to properly pull the input data based on a given sales val `run_id`.

While we are here, we might also want to do a very small migration and drop a recent sales val run's md5 hash for its input data into the `sale.metadata` table.

PR #169, opened 2025-11-19
Add dvc5 input data hash to `sale.metadata` and make small migrations

wagnerlmichael:
### Goals of this PR:
- Add code that persists the md5 hash from `dvc.lock` in the `sale.metadata` table. This was a missing link from https://github.com/ccao-data/model-sales-val/pull/164 that allows us to pull input data used by a given run through dvc by using the new column `dvc_md5_sales_ingest` to navigate the dvc bucket structure.
- Migration in  `sale.metadata`: Edit `2025-11-17_16:15-blissful-billy.parquet` to include the dvc md5 hash for the input data. We do this since this was a run that was completed after https://github.com/ccao-data/model-sales-val/pull/164 with dvc integration completed, and we ran `dvc push` after the run was complete.
- Migration in `sale.parameter` which catches a small inconsistency from https://github.com/ccao-data/model-sales-val/pull/165. In the data migration from this PR we erroneously included the top level key in the `standard_deviation_bounds` column. The sales val run `2025-11-17_16:15-blissful-billy` does not have this problem, since the code we implemented into the pipeline write the data correctly. But we found that `2025-11-17_16:15-blissful-billy` does have single quotes in the `standard_deviation_bounds` column where we need double quotes for proper json parsing. [We made an issue]( https://github.com/ccao-data/model-sales-val/issues/170) because it seems most of `sale.parameter` has this problem

### Migration nuance
- For `sale.metadata` we only need to delete and replace `2025-11-17_16:15-blissful-billy.parquet`

### Relevant S3 locations
Migration directory: `0005_retroactively_add_dvc_md5_hash_and_fix_sd_bounds/`

Saved prod data and updated data to migrate paths:
- `metadata_prior/`
- `metadata_updated/`
- `parameter_prior/`
- `parameter_updated/`

### Live data testing for the md5 persistence
- `SELECT * FROM "z_dev_miwagne_sale"."metadata"`

Closes https://github.com/ccao-data/model-sales-val/issues/168

### DIFF
--- a/dvc.lock
+++ b/dvc.lock
@@ -13,89 +13,89 @@ stages:
       size: 540
     - path: src/inputs.yaml
       hash: md5
-      md5: b1025386a92417a5ef58730cca961ad1
-      size: 4774
+      md5: 52a7d0a4834fcaef9803c443db6f3c56
+      size: 4802
     - path: src/utils.py
       hash: md5
-      md5: 0608bb585e246557c6cb151cf1839cec
-      size: 21658
+      md5: 3e360a2cfb99d1578ff0930883f9ba35
+      size: 21770
     outs:
     - path: input/sales_ingest.parquet
       hash: md5
-      md5: eaadf528757e2d3461bc0a5cc49e9d0c
-      size: 9302981
+      md5: 2a7ba5cfc8a00a47ab20ec3d313ea886
+      size: 9271421
   flag:
     cmd: uv run --env-file .env python3 src/01_flag.py
     deps:
     - path: input/sales_ingest.parquet
       hash: md5
-      md5: eaadf528757e2d3461bc0a5cc49e9d0c
-      size: 9302981
+      md5: 2a7ba5cfc8a00a47ab20ec3d313ea886
+      size: 9271421
     - path: src/01_flag.py
       hash: md5
-      md5: b84b8adb751de6b82fc7105d8c689a57
-      size: 12660
+      md5: 3006e84f14b22c039603e332c22da54d
+      size: 12656
     - path: src/constants.py
       hash: md5
       md5: 7c69d793d7e5917f4d6413316e64166f
       size: 540
     - path: src/inputs.yaml
       hash: md5
-      md5: b1025386a92417a5ef58730cca961ad1
-      size: 4774
+      md5: 52a7d0a4834fcaef9803c443db6f3c56
+      size: 4802
     - path: src/model.py
       hash: md5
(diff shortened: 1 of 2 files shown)


migrations/0005_retroactively_add_dvc_md5_hash_and_fix_sd_bounds.py
```diff
@@ -0,0 +1,80 @@
...
+import awswrangler as wr
+
+# sale.parameter migration
+# important: we are not touching the 2025-11-17_16:15-blissful-billy.parquet file in sale.parameter
+parquet_files_prod_prior = wr.s3.list_objects(
+    os.path.join(
+        "s3://ccao-data-backup-us-east-1/0005_retroactively_add_dvc_md5_hash_and_fix_sd_bounds/parameter_prior/"
+    ),
+    suffix=".parquet",
+)
```
jeancochrane: **[Suggestion, non-blocking]** I wonder if we should explicitly exclude `2025-11-17_16:15-blissful-billy.parquet`, similar to how we explicitly read it when updating `sale.metadata` below? It doesn't strike me as a huge deal because these migrations aren't fully reproducible anyway, but it would be slightly more accurate.

src/constants.py
```diff
@@ -12,3 +12,6 @@
 
 with open(os.path.join("src", "inputs.yaml"), "r") as stream:
     INPUTS = yaml.safe_load(stream)
+
+with open("dvc.lock", "r") as stream:
+    DVC_LOCKFILE = yaml.safe_load(stream)
```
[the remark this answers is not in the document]
wagnerlmichael: This makes sense to me, I'll move it out of the constants file. 

> I'm also curious: If you run a full dvc repro, will DVC update the lockfile after 00_ingest runs but before 01_flag runs?

I believe the answer is yes because after each stage finishes we see this message in the logs:
`Updating lock file 'dvc.lock'`

migrations/0005_retroactively_add_dvc_md5_hash_and_fix_sd_bounds.py
wagnerlmichael: Technically I didn't copy that file into that `parameter_prior` bucket, so it won't be there to begin with. Do you think we should still write that logic in here?

I guess maybe the most correct way to have done this would be to have programmatically copied the prod data into the backup bucket, and then make the exclusion there.

What do you think?

jeancochrane (approved):
Great work!

PR #172, opened 2026-01-18
Adjust default sales val south tri stat groupings to use custom defined nbhd groups

wagnerlmichael:
Editing the input spec so that we set the[ recently custom built geographic groups](https://github.com/ccao-data/data-architecture/pull/971) to be the default for residential south tri flagging. These groups were built in collaboration with res val.

### DIFF
--- a/src/inputs.yaml
+++ b/src/inputs.yaml
@@ -4,7 +4,6 @@
 # Note included with each run. Use this to summarize what changed about the run
 # or add context
 run_note: |
-  Testing full run of south tri nbhd_groups edit
 
 # 'dev' or 'prod'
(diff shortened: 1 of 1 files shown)


src/inputs.yaml
```diff
@@ -86,7 +86,7 @@ stat_groups:
     res_all:
       columns:
         - "rolling_window"
-        - "township_code"
+        - "geography_split"
```
wagnerlmichael: Only changes this for res. Leaning towards not changing for condos for the following reason
- Condos have very low sale volume in the south tri, and we are only reducing the group sizes with this geography change

PR #175, opened 2026-03-04
Remove group string references

### LINKED ISSUE #158: Remove group key column suffixes
In https://github.com/ccao-data/model-sales-val/pull/156 @wagnerlmichael discovered that the group key suffixes we append to our column names are generally not necessary, because each group gets evaluated separately by the modeling pipeline. As such, we should strip out these suffixes to reduce the complexity of the code.

wagnerlmichael:
In this PR I remove dynamically created 'group' suffixes on certain columns in `model.py`

The `create_group_string` function was a purely descriptive choice, and didn't have any functional impact on the pipeline. We can remove it safely because each group gets evaluated separately by the model script. None of these columns are persisted after the `model.py` script. In removing this, `model.py` should be ever so slightly more readable than it was before.

I did a test run locally and reviewed outputs, everything seems to be in order.

Closes https://github.com/ccao-data/model-sales-val/issues/158

### DIFF
--- a/src/model.py
+++ b/src/model.py
@@ -64,19 +64,6 @@ def go(
     return df
 
 
-def create_group_string(groups: tuple, sep: str) -> str:
-    """
-    Creates a string joined on a separator from the groups tuple.
-    For the purpose of making column names and descriptions.
-    Inputs:
-        groups (tuple): the columns being used in groupby()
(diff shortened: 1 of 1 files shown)


jeancochrane (approved):
Nice work! It's a small cleanup but I'm glad to have this resolved.

